PR #1, opened 2022-07-10
Proof of existence

bjaanes:
This PR is where the first module is added called prooforexistence. It is a pretty simple module that simply allows for hashes to be saved with a timestamp on the blockchain.

See the spec for more details on how it works if you are interested.


x/proofofexistence/handler.go
```diff
@@ -0,0 +1,29 @@
...
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/empowerchain/empowerchain/x/proofofexistence/keeper"
+	"github.com/empowerchain/empowerchain/x/proofofexistence/types"
+)
+
+// NewHandler ...
+func NewHandler(k keeper.Keeper) sdk.Handler {
```
marbar3778, line 13: these handlers are deprecated in 0.46. You can add them but I would suggest going straight to grpc and grpc-gateway for clients

marbar3778 (commented):
reviewed the code, haven't read the spec yest

PR #21, opened 2022-09-25
Preparing for the altruistic-1 testnet

nodersteam:
- Added path with testnets.
- Implemented a checker gentx for altruistic-1


bjaanes (approved):

.github/workflows/testnets.yml
```diff
@@ -0,0 +1,36 @@
+name: CI
+on:
+  pull_request:
+    branches: [ "main" ]
+    paths:
+      - 'altruistic-1/gentx/**.json'
```
bjaanes, line 6: This probably needs to be 'testnets/altruistic-1/gentx/**.json'

.github/workflows/testnets.yml
```diff
@@ -0,0 +1,36 @@
+name: CI
```
bjaanes, line 1: Let's change the name to Gentx checker or something?

.github/workflows/testnets.yml
icodragon: Corrected the path, thanks

.github/workflows/testnets.yml
icodragon: I like the name gentx checker - let's use it?

PR #166, opened 2022-10-02
Update README.md

Liver-23:
Set "Description=EmpowerChain Node" instead of "Description=Haqq Node"


bjaanes:
Thank you so much for that @Liver-23

PR #334, opened 2022-11-23
Access Control

ba1ciu:


chain/app/app.go
```diff
@@ -390,6 +396,7 @@ func New(
 		keys[plasticcreditmoduletypes.StoreKey],
 		keys[plasticcreditmoduletypes.MemStoreKey],
 		app.GetSubspace(plasticcreditmoduletypes.ModuleName),
+		app.GetPermStore(accesscontrolmoduletypes.ModuleName),
```
bjaanes, line 399: This should send in the keeper, right?

chain/x/accesscontrol/types/perm_store.go
```diff
@@ -0,0 +1,74 @@
+package types
```
bjaanes, line 1: This file makes me a little uncomfortable, because it deals with storage and stores and stuff outside of the keeper (go)module. I think it would be more correct to get this into the keeper in some way and send the keeper to the modules who are going to use this.

chain/x/accesscontrol/keeper/keeper.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (k Keeper) Logger(ctx sdk.Context) log.Logger {
+	return ctx.Logger().With("module", fmt.Sprintf("x/%s", types.ModuleName))
+}
+
+func (k Keeper) PermStore(name string) types.PermStore {
+	_, ok := k.permStores[name]
+	if ok {
+		panic("perm store already occupied")
```
bjaanes, line 42: More returning of errors and less panics, wtyd?

bjaanes (changes requested):
Some thoughts on the design. I like the general concept, but I would prefer to keep it more in line with typical SDK design of passing keepers and using interfaces (expected keepers type of thing)

bjaanes:
Oh, and spec documentation would be good :)

chain/x/accesscontrol/types/perm_store.go
ba1ciu: All storage op moved to keeper.

chain/x/accesscontrol/keeper/keeper.go
ba1ciu: Moved to subkeeper's initSubStore(). I'd personally leave it like that as SubKeeper is ought to be initialized only during app initialization in app.go

chain/app/app.go
ba1ciu: Now SubKeeper is initialized and injected.

chain/app/test_helpers.go
```diff
@@ -181,3 +182,10 @@ func CreateRandomAccounts(accNum int) []sdk.AccAddress {
 
 	return testAddrs
 }
+
+func InitAccessControlSubKeepers(keeper *accesscontrolmodulekeeper.Keeper) *[]accesscontrolmodulekeeper.IAccessControlSubKeeper {
+	var sk []accesscontrolmodulekeeper.IAccessControlSubKeeper
+	sk = append(sk, accesscontrolmodulekeeper.NewSubKeeper(keeper, "mockmodule1"))
```
gjermundgaraba, line 188: Do they have to be initialized?
I'm just wondering if it is less error-prone if they are just lazily initialized?

chain/proto/empowerchain/accesscontrol/query.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+package empowerchain.accesscontrol;
+
+import "gogoproto/gogo.proto";
+import "google/api/annotations.proto";
+
+option go_package = "github.com/empowerchain/empowerchain/x/accesscontrol/types";
+
+service Query {
```
gjermundgaraba, line 9: Would it make more sense to just not have these empty proto files if they are not really needed?

chain/x/accesscontrol/keeper/keeper.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (k Keeper) revokeAccess(ctx sdk.Context, subKey string, account sdk.AccAddress, msgType string) {
+	store := k.kvStore(ctx, subKey)
+	key := types.PermissionStoreKey(account, msgType)
+
+	store.Delete(key)
+}
+
+// Should be used only for genesis export!
+func (k Keeper) IteratePermissions(ctx sdk.Context, subKey string,
```
gjermundgaraba, line 67: If this is only for genesis export, can it not be private?

chain/x/accesscontrol/keeper/sub_keeper.go
```diff
@@ -0,0 +1,47 @@
...
+type IAccessControlSubKeeper interface {
+	HasAccess(ctx sdk.Context, account sdk.AccAddress, msgType string) bool
+	GrantAccess(ctx sdk.Context, account sdk.AccAddress, msgType string) error
+	RevokeAccess(ctx sdk.Context, account sdk.AccAddress, msgType string)
+}
+
+type SubKeeper struct {
+	*Keeper
+	subKey string
+	kaddr  string
```
gjermundgaraba, line 14: These names are very hard to parse and understand. Perhaps it might be better to give them more descriptive names?

chain/x/accesscontrol/module.go
```diff
@@ -0,0 +1,162 @@
...
+
+func (AppModuleBasic) RegisterLegacyAminoCodec(*codec.LegacyAmino) {}
+
+// RegisterInterfaces registers the module's interface types
+func (AppModuleBasic) RegisterInterfaces(registry codectypes.InterfaceRegistry) {
+	types.RegisterInterfaces(registry)
+}
+
+func (AppModuleBasic) DefaultGenesis(cdc codec.JSONCodec) json.RawMessage {
+	return nil
```
gjermundgaraba, line 58: Is nil valid here?

chain/x/accesscontrol/types/keys.go
```diff
@@ -0,0 +1,42 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/address"
+	"github.com/cosmos/cosmos-sdk/types/kv"
+)
+
+const (
+	// ModuleName defines the module name
+	ModuleName = "accesscontrol"
+
+	// Workaround for key collision
```
gjermundgaraba, line 15: Can you explain this a bit more?

gjermundgaraba (changes requested):
Looks good, mostly just some questions we can go through on Monday

chain/proto/empowerchain/accesscontrol/query.proto
ba1ciu: Yes. I deleted protos and querier + msg server.

chain/x/accesscontrol/keeper/keeper.go
ba1ciu: Fixed

chain/x/accesscontrol/keeper/sub_keeper.go
ba1ciu: Leftover from debugging. Deleted.

chain/x/accesscontrol/module.go
ba1ciu: Fixed.

chain/x/accesscontrol/types/keys.go
ba1ciu: Added a comment.

gjermundgaraba (changes requested):
Conflicts that needs to be solved, but otherwise good

PR #343, opened 2022-11-29
Issuer queries

gjermundgaraba:


chain/x/plasticcredit/keeper/issuer.go
```diff
@@ -15,11 +18,35 @@ func (k Keeper) GetIssuer(ctx sdk.Context, id uint64) (plasticcredit.Issuer, err
 	var issuer plasticcredit.Issuer
 	key := createKey(id)
 	bz := store.Get(key)
+	if len(bz) == 0 {
+		return issuer, errors.Wrapf(plasticcredit.ErrIssuerNotFound, "issuer with id: %d was not found", id)
```
ba1ciu, line 22: Maybe we could return 3rd value _found_ instead of returning the error?

gjermundgaraba: I was considering which one to use, happy to do this as well. Should we return an error in the querier itself?

gjermundgaraba: I like returning an error, similar to a 404. I would argue in most scenarios something is wrong if you look for a specific ID and you can't find it. But I am not 100% convinced this is the right way though :D

gjermundgaraba: I'll change it to "found"

ba1ciu (approved):

PR #352, opened 2022-12-02
Create credit class

gjermundgaraba:


gjermundgaraba:
@ba1ciu I'm just merging this in now so that I can avoid a lot of weird conflicts with the next task. Please still review this and I can fix it after-the-fact.

PR #375, opened 2022-12-08
360 update issuer

maheshKRex:
with guide from @gjermundgaraba 
added CLI for update-issuer and introduced event to trigger update-issuer


chain/proto/empowerchain/plasticcredit/tx.proto
```diff
@@ -17,6 +17,9 @@ service Msg {
   // There is only a single signer in the system that is allowed to create new issuers and it is defined in Params as "issuer_creator"
   rpc CreateIssuer(MsgCreateIssuer) returns (MsgCreateIssuerResponse);
 
+  // UpdateIssuer TODO: DOCUMENT
```
gjermundgaraba, line 20: Can you just write a single sentence on what this does (see the other ones for inspo)

chain/scripts/test/smoke_plastic_credit.sh
```diff
@@ -26,6 +26,11 @@ if [ "$NUM_ISSUERS" != "1" ]; then
   exit 1
 fi
 
+echo "--- Test: Plastic Credit Create Issuer ---"
```
gjermundgaraba, line 29:
```suggestion
echo "--- Test: Plastic Credit Update Issuer ---"
```

chain/x/plasticcredit/client/cli/tx.go
```diff
@@ -74,6 +75,52 @@ Note, the '--from' flag is ignored as it is implied from [admin_key_or_address].
 	return cmd
 }
 
+func MsgUpdateIssuerCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "update-issuer [admin_key_or_address] [issuer-id] [name] [description] ",
```
gjermundgaraba, line 80: I think it would be better here to just use the --from flag instead of having to specify the admin. 
So just update-issuer [issuer-id] [name] [description]

chain/x/plasticcredit/client/cli/tx.go
```diff
@@ -74,6 +75,52 @@ Note, the '--from' flag is ignored as it is implied from [admin_key_or_address].
...
+		Short: "Update existing Issuer.",
+		Long: `Update existing Issuer.
+Note, the '--from' flag is ignored as it is implied from [admin_key_or_address].
+`,
+		Args: cobra.MinimumNArgs(3),
```
gjermundgaraba, line 85: Let's use exact 3 (issuer, name and description)

chain/x/plasticcredit/client/cli/tx.go
```diff
@@ -74,6 +75,52 @@ Note, the '--from' flag is ignored as it is implied from [admin_key_or_address].
...
+			}
+
+			admin := clientCtx.GetFromAddress()
+			issuerId, err := cast.ToUint64E(args[1])
+			if err != nil {
+				return err
+			}
+			name := args[2]
+			var desc string
+			if len(args) > 3 {
```
gjermundgaraba, line 103: We should make description mandatory (even if you put in ""). Too easy to make a mistake here.
So then you can just remove the if check and just do `desc := args[3]`

gjermundgaraba (changes requested):

chain/proto/empowerchain/plasticcredit/tx.proto
```diff
@@ -17,7 +17,8 @@ service Msg {
   // There is only a single signer in the system that is allowed to create new issuers and it is defined in Params as "issuer_creator"
   rpc CreateIssuer(MsgCreateIssuer) returns (MsgCreateIssuerResponse);
 
-  // UpdateIssuer TODO: DOCUMENT
+  // UpdateIssuer updates an existing issuer details.
+  // There is only a single signer in the system that is allowed to update issuers and it is defined in Params as "updater"
```
gjermundgaraba, line 21:
```suggestion
  // The only entity that is allowed to change an issue is the address defined in the issuer as "admin"
```

gjermundgaraba (approved):

PR #378, opened 2022-12-09
Events - issuance/transfer/retire

ba1ciu:


chain/proto/empowerchain/plasticcredit/events.proto
```diff
@@ -21,4 +21,26 @@ message EventUpdateIssuer {
   string name = 3;
   string description = 4;
   string admin = 5 [(cosmos_proto.scalar) = "cosmos.AddressString"];
-}
\ No newline at end of file
+}
+
+// EventIssuedCredits is an event emitted when new credits are issued
+message EventIssuedCredits {
```
gjermundgaraba, line 27: I was thinking if we should actually also output some more info so that events are more useful for filtering and so on:
- credit class abbreviation
- issuer_id
- address that performed the issuance (not sure if this is actually useful, but the two other I think can be used in a good way for finding txs and stuff)

gjermundgaraba: The same goes for the other ones below

gjermundgaraba (changes requested):
Just a question if we should add some more details for the events to be more useful for filtering and search

chain/proto/empowerchain/plasticcredit/events.proto
ba1ciu: I'm not sure about this one. If we're gonna subscribe to all events and store them in some kind of the database, then querying, e.g. json-structured data should be simple enough and for credit class abbreviation and issuer id we can just:
- query denom event field with _abbrev/*_
- same as above but for all abbrevs for a given issuer

Or do you have something else in mind?

gjermundgaraba: The scenario I have in mind is actually to be able to query the chain for transactions based on events. The `txs` query takes an events today in the form of eventname.attributename=attributevalue. 
If we have have a bit more info in the events we could easily fetch transactions based on who issued them etc.

ba1ciu: Got you. Makes sense.

gjermundgaraba (approved):

PR #380, opened 2022-12-09
370 update applicant implementation, tests, cli

maheshKRex:


chain/proto/empowerchain/plasticcredit/tx.proto
```diff
@@ -93,6 +96,18 @@ message MsgCreateApplicantResponse{
   uint64 applicant_id = 1;
 }
 
+message MsgUpdateApplicant {
+  option (cosmos.msg.v1.signer) = "admin";
```
gjermundgaraba, line 100: The signer can be different from the new admin.

Do like the issuer update and have an updater signer.

chain/proto/empowerchain/plasticcredit/tx.proto
```diff
@@ -93,6 +96,18 @@ message MsgCreateApplicantResponse{
...
+  uint64 applicant_id = 1;
+  string name = 2;
+  string description = 3;
+  string admin = 4 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+}
+
+message MsgUpdateApplicantResponse{
+  uint64 applicant_id = 1;
```
gjermundgaraba, line 108: We don't really need to return anything here.

chain/x/plasticcredit/client/cli/tx.go
```diff
@@ -75,6 +76,48 @@ Note, the '--from' flag is ignored as it is implied from [admin_key_or_address].
...
+func MsgUpdateApplicantCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "update-applicant [admin_key_or_address] [applicant ID] [name] [description]",
+		Short: "Update existing applicant.",
+		Long: `Update existing applicant.
+Note, the '--from' flag is ignored as it is implied from [admin_key_or_address].
+`,
+		Args: cobra.MinimumNArgs(3),
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			if err := cmd.Flags().Set(flags.FlagFrom, args[0]); err != nil {
```
gjermundgaraba, line 88: Also here you will need to differentiate between admin and --from which might not be the same. So --from will be the "updater".

I do see that this is not set up correctly in update issuer cli also

chain/x/plasticcredit/keeper/applicant.go
```diff
@@ -42,7 +43,34 @@ func (k Keeper) createApplicant(ctx sdk.Context, name string, description string
...
-	return nextID, nil
+	return nextID, ctx.EventManager().EmitTypedEvent(&plasticcredit.EventCreateApplicant{
+		ApplicantId: applicant.Id,
+		Name:        applicant.Name,
+		Description: applicant.Description,
+		Admin:       applicant.Admin,
+	})
+}
+
+func (k Keeper) updateApplicant(ctx sdk.Context, applicantId uint64, name string, description string, admin string) error {
```
gjermundgaraba, line 54: This needs to take in the updater and check if the updater is an admin on the applicant.

chain/x/plasticcredit/keeper/msg_server_test.go
```diff
@@ -309,6 +310,82 @@ func (s *TestSuite) TestCreateApplicant() {
...
+		"happy path": {
+			msg: &plasticcredit.MsgUpdateApplicant{
+				ApplicantId: s.sampleApplicantId,
+				Name:        "Empower",
+				Description: "Empower is cool",
+				Admin:       sample.AccAddress(),
+			},
+			err: nil,
+		},
+		"applicant not found": {
```
gjermundgaraba, line 344: When you add the updater, make sure to add a test to see that only the admin of the applicant (s.sampleApplicantAdmin or something propbably) is allowed to do it.

gjermundgaraba (changes requested):

gjermundgaraba:
So a few comments and also check the linter and test. You can run them locally with `$ make lint`, `$ make test` and `$make smoketest`

chain/proto/empowerchain/plasticcredit/tx.proto
maheshKRex: added updater signer

chain/proto/empowerchain/plasticcredit/tx.proto
maheshKRex: fixed

chain/x/plasticcredit/client/cli/tx.go
maheshKRex: added updater signer, pls check and i'll update update issuer similarly

chain/x/plasticcredit/keeper/applicant.go
maheshKRex: added updater signer

chain/x/plasticcredit/keeper/applicant.go
```diff
@@ -42,7 +44,39 @@ func (k Keeper) createApplicant(ctx sdk.Context, name string, description string
...
-	return nextID, nil
+	return nextID, ctx.EventManager().EmitTypedEvent(&plasticcredit.EventCreateApplicant{
+		ApplicantId: applicant.Id,
+		Name:        applicant.Name,
+		Description: applicant.Description,
+		Admin:       applicant.Admin,
+	})
+}
+
+func (k Keeper) updateApplicant(ctx sdk.Context, applicantID uint64, name string, description string, admin string, updater sdk.AccAddress) error {
```
gjermundgaraba, line 55: Also not a big deal: but to keep it similar to the other ones I prefer that the updater address comes in as the first argument (after ctx).

so it would be `updateApplicant(ctx, updater, applicantID, etc`

gjermundgaraba (changes requested):
Just some small things :)

chain/x/plasticcredit/keeper/applicant.go
gjermundgaraba: This has not been updated yet?

chain/x/plasticcredit/keeper/msg_server_test.go
```diff
@@ -309,6 +310,105 @@ func (s *TestSuite) TestCreateApplicant() {
...
+			msg: &plasticcredit.MsgUpdateApplicant{
+				Updater:     sample.AccAddress(), // not allowed!
+				ApplicantId: 1,
+				Name:        "Empower",
+				Description: "Empower is cool",
+				Admin:       sample.AccAddress(),
+			},
+			err: sdkerrors.ErrUnauthorized,
+		},
+		"invalid address": {
```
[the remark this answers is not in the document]
gjermundgaraba: This is also missing :)

gjermundgaraba (changes requested):
A couple of things not fixed yet

PR #395, opened 2022-12-16
reject project implementation, tests, validations, updates to issuer tests

maheshKRex:


chain/x/plasticcredit/keeper/msg_server_test.go
```diff
@@ -836,14 +843,110 @@ func (s *TestSuite) TestApproveProject() {
...
+			err: sdkerrors.ErrUnauthorized,
+		},
+		"project not found": {
+			msg: &plasticcredit.MsgRejectProject{
+				Rejector:  s.sampleIssuerAdmin,
+				ProjectId: 42,
+			},
+			err: plasticcredit.ErrProjectNotFound,
+		},
+		"project already rejected": {
```
gjermundgaraba, line 905: test also project already approved project: It should not be possible to reject an already approved project.

chain/x/plasticcredit/keeper/project.go
```diff
@@ -122,6 +124,42 @@ func (k Keeper) ApproveProject(ctx sdk.Context, approver sdk.AccAddress, project
 	})
 }
 
+func (k Keeper) RejectProject(ctx sdk.Context, rejector sdk.AccAddress, projectID uint64) error {
+	project, found := k.GetProject(ctx, projectID)
+	if !found {
+		return errors.Wrapf(plasticcredit.ErrProjectNotFound, "project with id %d was not found", projectID)
+	}
+	if project.Status == plasticcredit.ProjectStatus_REJECTED {
```
gjermundgaraba, line 132: I think we can just check directly that the project is in the state NEW since it should also not be allowed to reject an already approved project.

gjermundgaraba (changes requested):
Really good! Just one business rule that I believe is not implemented: It should not be possible to reject an already approved project

ba1ciu (approved):

chain/x/plasticcredit/keeper/project.go
maheshKRex: yeah, good point. fixed.

gjermundgaraba (approved):

Issue #382, opened 2022-12-09
Combine genesis test with invariant check


gjermundgaraba:

ba1ciu:
@gjermundgaraba 

My idea of reusing stuff we already implemented for genesis export would be to refactor getting all store records to iterators that use callback/handler functions like [we have in accesscontrol](https://github.com/EmpowerPlastic/empowerchain/blob/91ae1e730175243e08a2db940efd565c84ef7ed7/chain/x/accesscontrol/keeper/keeper.go#L73). Then we can just pass a handler that creates a slice of all objects in case of export and a different handler that checks invariant for invariant.go.

gjermundgaraba:
Sounds perfect! LFG!

ba1ciu:
@gjermundgaraba @maheshKRex let's list below what we think should be a system invariant for plastic credit

ba1ciu:
1. Sum of all credit balances should equal a total number of credits in the credit collection.
2. ID Counters should point to next unexisting ID.

gjermundgaraba:
I think we can break down the balance checks more too to also make sure the total number of credits within a denom is correct (balances.sum == collections.sum). Then a final total check.

Can't really think of anything else that would should be chain-halting. I assume if any links are broken for instance, the worst that will happen is that entities can't function (e.g. a project with a broken applicant_id link would just not be possible to use, right?), not that they would open any potential security vulns.

ba1ciu:
Exactly. It will break using state that is corrupted in any way and shouldn't have any other consequences

gjermundgaraba:
@ba1ciu Where are we on this one?

ba1ciu:
So as far as I remember, when we discussed the matter with Marko, we agreed that only chain halting events business-wise should be implemented as invariants. I don't think there are any others than credit balance corruption?

gjermundgaraba:
Perhaps also the next counters should not point to an existing entity? If it does I also think something is very bad and things can get corrupted?

ba1ciu:
I thought about that, but wasn't sure if we actually need that as counters are so simple that it's going to be really hard to break them? Anyway, this invariant check won't be expensive so we can as well add it.

gjermundgaraba:
True, but better to be on the safe side. This way, if we do mess them up, they would more easily be caught in simulation tests too.

gjermundgaraba:
So let's do that as well in this issue, and then we can close it

ba1ciu:
[...]

PR #397, opened 2022-12-16
Invariants

ba1ciu:


gjermundgaraba:
@ba1ciu this PR looks like it does not cover: https://github.com/EmpowerPlastic/empowerchain/issues/382 fully

If I read the code correctly we still don't have all the tests in the genesis validation incorporated into invariant checking? Will that be a separate PR (which is fine, but then we need to just unlink 382 from this pr)

ba1ciu:
@gjermundgaraba Yes, the idea was to merge the base for invariant checking and add the rest later. I unpinned #382.

While we're at it, if we want to use all our genesis validation for invariant checking, which is e.g. checking if credit collection's project id exists in the store, we have to keep in mind that if for some reason state gets corrupted and invariant fails, the chain is halted, so maybe it's not a good idea to make every genesis validation an invariant?

gjermundgaraba:
> if for some reason state gets corrupted

True, you're right that we need to be careful about it, but at the same time, state corruption sounds like a potentially good reason to halt?

chain/x/plasticcredit/keeper/credit.go
```diff
@@ -298,34 +298,34 @@ func (k Keeper) setCreditBalance(ctx sdk.Context, balance plasticcredit.CreditBa
...
 
 	iterator := store.Iterator(nil, nil)
 	defer iterator.Close()
 
-	var creditCollections []plasticcredit.CreditCollection
 	for ; iterator.Valid(); iterator.Next() {
 		var creditCollection plasticcredit.CreditCollection
 		k.cdc.MustUnmarshal(iterator.Value(), &creditCollection)
-		creditCollections = append(creditCollections, creditCollection)
+		if handler(creditCollection) {
```
gjermundgaraba, line 310: The same as the other comment on handler, something with naming here to help understand the return value. Either a better name for the `handler` or name the function `type goodName func....`

chain/x/plasticcredit/keeper/credit.go
```diff
@@ -298,34 +298,34 @@ func (k Keeper) setCreditBalance(ctx sdk.Context, balance plasticcredit.CreditBa
...
 
 	iterator := store.Iterator(nil, nil)
 	defer iterator.Close()
 
-	var creditBalances []plasticcredit.CreditBalance
 	for ; iterator.Valid(); iterator.Next() {
 		var creditBalance plasticcredit.CreditBalance
 		k.cdc.MustUnmarshal(iterator.Value(), &creditBalance)
-		creditBalances = append(creditBalances, creditBalance)
+		if handler(creditBalance) {
```
gjermundgaraba, line 325: Same

chain/x/plasticcredit/keeper/credit_class.go
```diff
@@ -121,20 +121,19 @@ func (k Keeper) setCreditClass(ctx sdk.Context, creditClass plasticcredit.Credit
...
 
 	iterator := store.Iterator(nil, nil)
 	defer iterator.Close()
 
-	var creditClasses []plasticcredit.CreditClass
 	for ; iterator.Valid(); iterator.Next() {
 		var creditClass plasticcredit.CreditClass
 		k.cdc.MustUnmarshal(iterator.Value(), &creditClass)
-		creditClasses = append(creditClasses, creditClass)
+		if handler(creditClass) {
```
gjermundgaraba, line 133: +1

chain/x/plasticcredit/keeper/invariants.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func AllInvariants(k Keeper) sdk.Invariant {
+	return func(ctx sdk.Context) (string, bool) {
+		return TotalSupply(k)(ctx)
+	}
+}
+
+// TotalSupply checks that the total supply reflects all the credits held in accounts
+func TotalSupply(k Keeper) sdk.Invariant {
```
gjermundgaraba, line 22: Should probably be renamed to something like TotalSupplyInvariant? The name, which is on the keeper, reads like it returns the total supply.

chain/x/plasticcredit/keeper/invariants_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"github.com/EmpowerPlastic/empowerchain/testutil/sample"
+	"github.com/EmpowerPlastic/empowerchain/x/plasticcredit"
+	"github.com/EmpowerPlastic/empowerchain/x/plasticcredit/keeper"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func (s *TestSuite) TestTotalSuppliesInvariant() {
+	s.Run("happy path", func() {
```
gjermundgaraba, line 11: Hmm, it would be best if we managed to find a way to actually test an invariant. Might require this package to be `keeper` to get access to low-level setter methods perhaps?

chain/x/plasticcredit/keeper/issuer.go
```diff
@@ -45,20 +45,19 @@ func (k Keeper) GetIssuers(ctx sdk.Context, pageReq query.PageRequest) ([]plasti
...
 
 	iterator := store.Iterator(nil, nil)
 	defer iterator.Close()
 
-	var issuers []plasticcredit.Issuer
 	for ; iterator.Valid(); iterator.Next() {
 		var issuer plasticcredit.Issuer
 		k.cdc.MustUnmarshal(iterator.Value(), &issuer)
-		issuers = append(issuers, issuer)
+		if handler(issuer) {
```
gjermundgaraba, line 57: +1

chain/x/plasticcredit/keeper/project.go
```diff
@@ -139,20 +139,19 @@ func (k Keeper) setProject(ctx sdk.Context, project plasticcredit.Project) error
...
 
 	iterator := store.Iterator(nil, nil)
 	defer iterator.Close()
 
-	var projects []plasticcredit.Project
 	for ; iterator.Valid(); iterator.Next() {
 		var project plasticcredit.Project
 		k.cdc.MustUnmarshal(iterator.Value(), &project)
-		projects = append(projects, project)
+		if handler(project) {
```
gjermundgaraba, line 151: +1

gjermundgaraba (changes requested):
Looks good! A couple of naming things and some more tests.

chain/x/plasticcredit/keeper/applicant.go
```diff
@@ -96,20 +96,19 @@ func (k Keeper) setApplicant(ctx sdk.Context, applicant plasticcredit.Applicant)
...
 
 	iterator := store.Iterator(nil, nil)
 	defer iterator.Close()
 
-	var applicants []plasticcredit.Applicant
 	for ; iterator.Valid(); iterator.Next() {
 		var applicant plasticcredit.Applicant
 		k.cdc.MustUnmarshal(iterator.Value(), &applicant)
-		applicants = append(applicants, applicant)
+		if handler(applicant) {
```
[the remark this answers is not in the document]
gjermundgaraba: As per discussion, we will remove the bool return value from the handler as it is not used right now.

gjermundgaraba (approved):

Issue #396, opened 2022-12-16
Suspending projects


gjermundgaraba:
Currently the current business rule does not allow for reverting an approval of a project "It should not be possible to reject an already approved project".

To make this a bit more flexible and usable, we should add the possibility to SUSPEND an already approved project, even if it has already been issued credits.

It should also be possible to approve a project again from the suspended state.

- [x] Add new SUSPENDED state to projects
- [x] Create new message to suspend a project
- [ ] Change approve project message logic to also allow approving a suspended project
- [x] Add tests to cover the new suspended state when issuing credits
- [x] Add tests to cover the new suspended state when rejecting a project
- [ ] Add tests to cover approving a suspended project works

Business rules/state machine:
- A project is first in the NEW state
- A project can be be put in the APPROVED state if, and only if, it is in the NEW or SUSPENDED state (in other words, not when REJECTED)
- A project can be put in the REJECTED state if, and only if, it is in the NEW state (in other words, not from APPROVED or SUSPENDED)
- A project can suspended if, and only if, it is in the APPROVED state (in other words, not when in NEW or REJECTED)
- Credits can be issued if, and only if, the project is in the APPROVED state (in other words, not when in NEW, SUSPENDED or REJECTED)

```mermaid
stateDiagram-v2
    NEW --> APPROVED
    NEW --> REJECTED
    APPROVED --> SUSPENDED
    SUSPENDED --> APPROVED
```

PR #401, opened 2022-12-21
suspend project capability, tests, validaitons

maheshKRex:
https://github.com/EmpowerPlastic/empowerchain/issues/396


gjermundgaraba (changes requested):
The only thing we need to add a test for is that is possible to approve a suspended project. Add that test to the approve project tests.

maheshKRex:
> The only thing we need to add a test for is that is possible to approve a suspended project. Add that test to the approve project tests.

added tests, and also fixed existing ones to check correct project id in the test case.

gjermundgaraba (approved):

PR #405, opened 2022-12-22
e2e test suite

ba1ciu:


chain/tests/e2e/README.md
```diff
@@ -0,0 +1,26 @@
+# E2E tests
```
gjermundgaraba, line 1: This README is very nice!

chain/tests/e2e/plasticcredit/e2e_test.go
```diff
@@ -0,0 +1,246 @@
...
+	s.cfg.GenesisState = genesisState
+
+	s.cfg.AppConstructor = NewAppConstructor()
+
+	s.network, err = network.New(s.T(), s.T().TempDir(), s.cfg)
+	s.Require().NoError(err)
+
+	kb := s.network.Validators[0].ClientCtx.Keyring
+	_, err = kb.NewAccount(
+		"issuer",
```
gjermundgaraba, line 207: Since we use these names other places in tests, pull them up into constants.

chain/tests/e2e/plasticcredit/query_test.go
```diff
@@ -0,0 +1,89 @@
...
+	s.Require().NoError(err)
+	testCases := []struct {
+		name           string
+		args           []string
+		expectedErr    bool
+		expectedErrMsg string
+		expectedResp   proto.Message
+	}{
+		{
+			"query existing issuer",
```
gjermundgaraba, line 53: Just to keep our tests consistent, can you change from testCases being `[]struct { name string}` to our current `map[string]struct{}`? 

Here and in the other tests as well.

gjermundgaraba (changes requested):
This makes a lot of sense and looks really good. Just a couple of small details :)

gjermundgaraba (approved):

PR #414, opened 2022-12-28
e2e tests for project

maheshKRex:
work in progress for create project CLI smoketests, keeping in draft

### DIFF
--- a/chain/tests/e2e/plasticcredit/e2e_test.go
+++ b/chain/tests/e2e/plasticcredit/e2e_test.go
@@ -119,7 +119,7 @@ func (s *E2ETestSuite) SetupSuite() {
 			Id:                      2,
 			ApplicantId:             1,
 			CreditClassAbbreviation: "PCRD",
-			Name:                    "Also approved project",
+			Name:                    "Suspended project",
 			Status:                  plasticcredit.ProjectStatus_SUSPENDED,
 		},
 		{
--- a/chain/tests/e2e/plasticcredit/e2e_test.go
+++ b/chain/tests/e2e/plasticcredit/e2e_test.go
@@ -71,7 +71,7 @@ func (s *E2ETestSuite) SetupSuite() {
 	plasticcreditGenesisState.IdCounters = plasticcredit.IDCounters{
 		NextIssuerId:    3,
 		NextApplicantId: 2,
-		NextProjectId:   4,
+		NextProjectId:   5,
 	}
 	plasticcreditGenesisState.Issuers = []plasticcredit.Issuer{
 		{
--- a/chain/tests/e2e/plasticcredit/e2e_test.go
+++ b/chain/tests/e2e/plasticcredit/e2e_test.go
@@ -71,7 +71,7 @@ func (s *E2ETestSuite) SetupSuite() {
 	plasticcreditGenesisState.IdCounters = plasticcredit.IDCounters{
 		NextIssuerId:    3,
 		NextApplicantId: 2,
-		NextProjectId:   5,
+		NextProjectId:   7,
 	}
 	plasticcreditGenesisState.Issuers = []plasticcredit.Issuer{
 		{
@@ -135,6 +135,20 @@ func (s *E2ETestSuite) SetupSuite() {
 			CreditClassAbbreviation: "PCRD",
 			Name:                    "Rejected project",
 			Status:                  plasticcredit.ProjectStatus_REJECTED,
+		},
+		{
+			Id:                      5,
+			ApplicantId:             1,
+			CreditClassAbbreviation: "PCRD",
+			Name:                    "Other New Project",
+			Status:                  plasticcredit.ProjectStatus_NEW,
+		},
+		{
+			Id:                      6,
+			ApplicantId:             1,
+			CreditClassAbbreviation: "PCRD",
+			Name:                    "Another New Project",
+			Status:                  plasticcredit.ProjectStatus_NEW,
 		},
 	}
 	plasticcreditGenesisState.CreditCollections = []plasticcredit.CreditCollection{
--- a/chain/tests/e2e/plasticcredit/e2e_test.go
+++ b/chain/tests/e2e/plasticcredit/e2e_test.go
@@ -120,7 +120,7 @@ func (s *E2ETestSuite) SetupSuite() {
 			ApplicantId:             1,
 			CreditClassAbbreviation: "PCRD",
 			Name:                    "Also approved project",
-			Status:                  plasticcredit.ProjectStatus_APPROVED,
+			Status:                  plasticcredit.ProjectStatus_SUSPENDED,
 		},
 		{
 			Id:                      3,
--- a/chain/tests/e2e/plasticcredit/tx_test.go
+++ b/chain/tests/e2e/plasticcredit/tx_test.go
@@ -143,7 +143,7 @@ func (s *E2ETestSuite) TestCmdCreateProject() {
 			false,
 			"",
 			&plasticcredit.Project{
-				Id:                      5,
+				Id:                      7,
 				ApplicantId:             1,
 				CreditClassAbbreviation: "EMP",
 				Name:                    "My new Project",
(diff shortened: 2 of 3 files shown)


chain/tests/e2e/plasticcredit/tx_test.go
```diff
@@ -123,3 +123,223 @@ func (s *E2ETestSuite) TestCmdUpdateIssuer() {
...
+			if tc.expectedErrOnSend {
+				s.Require().Contains(out.String(), tc.expectedErrMsg)
+			} else if tc.expectedErrOnExec {
+				var txResponse sdk.TxResponse
+				s.Require().NoError(val.ClientCtx.Codec.UnmarshalJSON(out.Bytes(), &txResponse))
+				s.Require().Contains(txResponse.RawLog, tc.expectedErrMsg)
+			} else {
+				var txResponse sdk.TxResponse
+				s.Require().NoError(val.ClientCtx.Codec.UnmarshalJSON(out.Bytes(), &txResponse))
+				fmt.Println(txResponse.RawLog)
```
maheshKRex, line 176: @ba1ciu this still fails to create a new project, even if i tried to use the admin address of the applicant, for --from flag. By looking at the genesis the new Project ID should be 5, but it doesn't find the project.
so i've been trying to trace the output and print RawLog, Body ...but they end up giving the marshalled strings, which i'm not sure unmarshal again like 123A0A342F656D706F776572636. pls help

ba1ciu: Let's figure it out on Slack.

chain/tests/e2e/plasticcredit/e2e_test.go
```diff
@@ -120,7 +120,7 @@ func (s *E2ETestSuite) SetupSuite() {
 			ApplicantId:             1,
 			CreditClassAbbreviation: "PCRD",
 			Name:                    "Also approved project",
-			Status:                  plasticcredit.ProjectStatus_APPROVED,
+			Status:                  plasticcredit.ProjectStatus_SUSPENDED,
```
ba1ciu, line 123: Since we're changing the status, can we also change the Name, so it shows that the project is suspended?

chain/tests/e2e/plasticcredit/tx_test.go
```diff
@@ -123,3 +123,358 @@ func (s *E2ETestSuite) TestCmdUpdateIssuer() {
...
+			},
+			"5",
+		},
+		"admin does not have authorization for applicant": {
+			[]string{"1", "EMP", "My new Project", fmt.Sprintf("--%s=%s", flags.FlagFrom, "empower1qnk2n4nlkpw9xfqntladh74w6ujtulwnz7rf8m")},
+			true,
+			false,
+			"unauthorized",
+			nil,
+			"6",
```
ba1ciu, line 163: Looks unnecessary here?

chain/tests/e2e/plasticcredit/tx_test.go
```diff
@@ -123,3 +123,358 @@ func (s *E2ETestSuite) TestCmdUpdateIssuer() {
...
+			false,
+			"",
+			&plasticcredit.Project{
+				Id:                      5,
+				ApplicantId:             1,
+				CreditClassAbbreviation: "EMP",
+				Name:                    "My new Project",
+				Status:                  plasticcredit.ProjectStatus_NEW,
+			},
+			"5",
```
ba1ciu, line 155: Not sure, but wouldn't it be more convenient to just use expected state's ID to query the project?

chain/tests/e2e/plasticcredit/tx_test.go
```diff
@@ -123,3 +123,358 @@ func (s *E2ETestSuite) TestCmdUpdateIssuer() {
...
+				s.Require().NoError(val.ClientCtx.Codec.UnmarshalJSON(out.Bytes(), &resp))
+				s.Require().Equal(tc.expectedState, &resp.Project)
+			}
+		})
+	}
+}
+
+func (s *E2ETestSuite) TestCmdRejectProject() {
+	val := s.network.Validators[0]
+	admin := "empower1qnk2n4nlkpw9xfqntladh74w6ujtulwnz7rf8m"
```
ba1ciu, line 346: I'd fetch it from the keyring by an issuerKey to avoid having raw address here. And same in other tests.

chain/tests/e2e/plasticcredit/tx_test.go
```diff
@@ -123,3 +123,358 @@ func (s *E2ETestSuite) TestCmdUpdateIssuer() {
...
+	}{
+		"invalid admin": {
+			[]string{"5", fmt.Sprintf("--%s=%s", flags.FlagFrom, notAdminKey.Name)},
+			false,
+			true,
+			"unauthorized",
+			nil,
+		},
+		"reject project": {
+			[]string{"5", fmt.Sprintf("--%s=%s", flags.FlagFrom, admin)},
```
ba1ciu, line 364: I'd avoid using objects created by other tests. We're not sure about the order the test are executed, so it's better to use objects created at genesis level.

chain/x/plasticcredit/keeper/project.go
```diff
@@ -147,10 +144,12 @@ func (k Keeper) RejectProject(ctx sdk.Context, rejector sdk.AccAddress, projectI
 	if !found {
 		panic("The credit class was found, the issuer better exist!")
 	}
-
 	if !issuer.AddressHasAuthorization(rejector) {
 		return errors.Wrapf(sdkerrors.ErrUnauthorized, "rejector %s does not have authorization on issuer with id %d", rejector.String(), issuer.Id)
 	}
+	if project.Status != plasticcredit.ProjectStatus_NEW {
```
ba1ciu, line 150: Is there a reason to move this check further down?

chain/x/plasticcredit/keeper/project.go
```diff
@@ -183,11 +179,12 @@ func (k Keeper) SuspendProject(ctx sdk.Context, updater sdk.AccAddress, projectI
 	if !found {
 		panic("The credit class was found, the issuer better exist!")
 	}
-
 	if !issuer.AddressHasAuthorization(updater) {
 		return errors.Wrapf(sdkerrors.ErrUnauthorized, "For suspension the updater %s does not have authorization on issuer with id %d", updater.String(), issuer.Id)
 	}
-
+	if project.Status != plasticcredit.ProjectStatus_APPROVED {
```
ba1ciu, line 185: As above.

ba1ciu (changes requested):

chain/tests/e2e/plasticcredit/tx_test.go
maheshKRex: yes, started using expectedState.Id

chain/tests/e2e/plasticcredit/tx_test.go
maheshKRex: yes, started using expectedState.Id

chain/tests/e2e/plasticcredit/tx_test.go
maheshKRex: updated to use issuerKey

chain/tests/e2e/plasticcredit/tx_test.go
maheshKRex: yeah. added 2 new projects to genesis and made tests independent

chain/x/plasticcredit/keeper/project.go
maheshKRex: it was i was just trying to test the authorisation check using approved/ rejected projects, instead started using fresh projects, and moved back the checks to where it was before.

ba1ciu:
> it was i was just trying to test the authorisation check using approved/ rejected projects, instead started using fresh projects, and moved back the checks to where it was before. 

The reason I was asking was that good practice is to perform checks as soon as possible, so failed transactions use least gas possible.

ba1ciu (approved):

PR #416, opened 2023-01-03
added e2e tests for create, update applicant

maheshKRex:
https://github.com/EmpowerPlastic/empowerchain/issues/408


chain/tests/e2e/plasticcredit/applicant_tx_test.go
```diff
@@ -0,0 +1,142 @@
...
+	s.Require().NoError(err)
+
+	testCases := map[string]struct {
+		args              []string
+		expectedErrOnSend bool
+		expectedErrOnExec bool
+		expectedErrMsg    string
+		expectedState     *plasticcredit.Applicant
+	}{
+		"create applicant": {
```
ba1ciu, line 26: We could add the test for empty name (error) and empty description (which should pass) and invalid admin address.

chain/tests/e2e/plasticcredit/project_tx_test.go
```diff
@@ -0,0 +1,364 @@
...
+	notApplicantKey, _ := val.ClientCtx.Keyring.Key(issuerKey)
+	testCases := map[string]struct {
+		args              []string
+		expectedErrOnSend bool
+		expectedErrOnExec bool
+		expectedErrMsg    string
+		expectedState     *plasticcredit.Project
+	}{
+
+		"create project": {
```
ba1ciu, line 26: Empty name and empty description?

chain/tests/e2e/plasticcredit/project_tx_test.go
```diff
@@ -0,0 +1,364 @@
...
+	s.Require().NoError(err)
+	testCases := map[string]struct {
+		args              []string
+		expectedErrOnSend bool
+		expectedErrOnExec bool
+		expectedErrMsg    string
+		expectedState     *plasticcredit.Project
+	}{
+
+		"approve project": {
```
ba1ciu, line 156:
```suggestion
		"approve new project": {
```
Just for clarification

chain/tests/e2e/plasticcredit/project_tx_test.go
```diff
@@ -0,0 +1,364 @@
...
+			nil,
+		},
+		"invalid admin": {
+			[]string{"4", fmt.Sprintf("--%s=%s", flags.FlagFrom, notAdminKey.Name)},
+			false,
+			true,
+			"unauthorized",
+			nil,
+		},
+		"project already approved": {
```
ba1ciu, line 196: Maybe let's add approving rejected project also?

chain/tests/e2e/plasticcredit/project_tx_test.go
```diff
@@ -0,0 +1,364 @@
...
+			nil,
+		},
+		"reject non-existing project": {
+			[]string{"62", fmt.Sprintf("--%s=%s", flags.FlagFrom, issuerKey.Name)},
+			false,
+			true,
+			"project not found",
+			nil,
+		},
+		"project already approved": {
```
ba1ciu, line 266: Rejecting already rejected and suspended?

chain/tests/e2e/plasticcredit/project_tx_test.go
```diff
@@ -0,0 +1,364 @@
...
+	notAdminKey, err := val.ClientCtx.Keyring.Key(applicantKey)
+	s.Require().NoError(err)
+	testCases := map[string]struct {
+		args              []string
+		expectedErrOnSend bool
+		expectedErrOnExec bool
+		expectedErrMsg    string
+		expectedState     *plasticcredit.Project
+	}{
+		"reject project": {
```
ba1ciu, line 239:
```suggestion
		"reject new project": {
```

chain/tests/e2e/plasticcredit/project_tx_test.go
```diff
@@ -0,0 +1,364 @@
...
+	notAdminKey, err := val.ClientCtx.Keyring.Key(applicantKey)
+	s.Require().NoError(err)
+	testCases := map[string]struct {
+		args              []string
+		expectedErrOnSend bool
+		expectedErrOnExec bool
+		expectedErrMsg    string
+		expectedState     *plasticcredit.Project
+	}{
+		"suspend project": {
```
ba1ciu, line 309:
```suggestion
		"suspend approved project": {
```

chain/tests/e2e/plasticcredit/project_tx_test.go
```diff
@@ -0,0 +1,364 @@
...
+			nil,
+		},
+		"suspending non-existing project": {
+			[]string{"62", fmt.Sprintf("--%s=%s", flags.FlagFrom, issuerKey.Name)},
+			false,
+			true,
+			"project not found",
+			nil,
+		},
+		"project already rejected": {
```
ba1ciu, line 336: Suspending already suspended and new projects?

chain/tests/e2e/plasticcredit/project_tx_test.go
```diff
@@ -0,0 +1,364 @@
...
+		expectedErrMsg    string
+		expectedState     *plasticcredit.Project
+	}{
+
+		"update project": {
+			[]string{"3", "My Updated Project", fmt.Sprintf("--%s=%s", flags.FlagFrom, applicantKey.Name)},
+			false,
+			false,
+			"",
+			&plasticcredit.Project{
```
ba1ciu, line 90: Project with ID 3 is used for both update and approve testing. You should use different for each case. We should keep in mind that the test should run successfully both as a whole suite and independently. If you run just this test, it will fail, because Project 3 is initially in new state.

ba1ciu (changes requested):
Couple suggestions about other test scenarios, mostly testing all possible transitions between project statuses and one small issue with project update to fix. Those are not specifically for the applicant case, so we can create separate issue for those or fix right away :)

chain/tests/e2e/plasticcredit/applicant_tx_test.go
maheshKRex: @ba1ciu i've been struggling with the invalid admin test for create applicant. Similarly to other invalid admin tests i tried to provide a test case with simply an other address string as the admin account address instead of issuer/applicant address from keyring. 
`"invalid admin address": {
			[]string{"empower1qnk2n4nlkpw9xfqntladh74w6ujtulwnz7rf", "New", "Grab that bottle", fmt.Sprintf("--%s=%s", flags.FlagFrom,issuerKey.Name)},
		`
But the clitestutil.ExecTestCLICmd fails to pass this string to the test with error key not found, so when invoking the CLI this way, does is always need to have a valid key AccAddress, like issuer / applicant key address?
`empower1qnk2n4nlkpw9xfqntladh74w6ujtulwnz7rf8.info: key not found`

maheshKRex: found that this is imposed by validation check at CLI x/plasticcredit/client/cli/tx.go forcing to take --from addr as the Admin, making it to always have a valid AccAddress. So it can be any address as we don't restrict anybody with a signing address to go ahead and create an applicant.
`if err := cmd.Flags().Set(flags.FlagFrom, args[0]); err != nil {
	return err
}`
So should we still do an invalid admin test, or just test for validity of from address.

chain/tests/e2e/plasticcredit/project_tx_test.go
maheshKRex: added.

chain/tests/e2e/plasticcredit/project_tx_test.go
maheshKRex: agree. updated to use new project for this. https://github.com/EmpowerPlastic/empowerchain/blob/702dbf9b8b0940edd69e170941155eba76c195ba/chain/tests/e2e/plasticcredit/project_tx_test.go#L104

chain/tests/e2e/plasticcredit/project_tx_test.go
maheshKRex: added.

chain/tests/e2e/plasticcredit/project_tx_test.go
maheshKRex: added.

chain/tests/e2e/plasticcredit/applicant_tx_test.go
ba1ciu: I think we should leave validation in tx.go as it is, so the cli works both with addresses and key names.
If the message signer is an admin and therefore you cannot give anyone other than yourself an admin role, then I suggest we either drop the admin param and take it from the --from flag or allow to assign admin to others by adding a creator field, which will hold the address from --from. Wdyt, @gjermundgaraba?

maheshKRex: yes, updated create-applicant to take admin using --from flag, and it looks much neater.

ba1ciu (approved):

PR #438, opened 2023-01-21
Fix whitespace for Make 4.3

freshe4qa:


gjermundgaraba:
Hi @freshe4qa! Thanks for the PR! Would you mind explaining what this change actually does and why it is needed? :)

freshe4qa:
Optimizing for the new version

gjermundgaraba:
@freshe4qa can you please be more specific? Optimizing how? Solving what problem? I just need to understand what I am merging here :)

freshe4qa:
Make 4.3 will only add a space when += is used if the variable it is used on exists. Using += on a non-existing variable would previously create a variable starting with a space.

gjermundgaraba (approved):

gjermundgaraba:
Thanks for this @freshe4qa :rocket:

freshe4qa:
Always glad to help

PR #442, opened 2023-01-23
393 upgrade to cosmos sdk 047

gjermundgaraba:
Sorry for the big PR. There was a lot of changes needed - including  the e2e tests which broke badly by the removal of the `block` broadcast-mode. `sync` does not work the same way unfortunately.

I also removed IBC for now since ibc-go doesn't support 0.47 yet (and depending on the main branch was a bad idea). Will be easy to add back soon when ibc-go 7 is released.


chain/Makefile
```diff
@@ -145,11 +145,15 @@ smoketest: install
 ###                                  Proto                                  ###
 ###############################################################################
 
+protoVer=0.11.2
```
aljo242, line 148: Newer version of the proto-builder container exists:

https://github.com/cosmos/cosmos-sdk/blob/fdaafa0f1e1bc2317932cf760266a3c1af3c9703/Makefile#L398

chain/tests/e2e/plasticcredit/e2e_test.go
```diff
@@ -3,6 +3,7 @@ package e2e_test
 import (
 	"encoding/hex"
 	"fmt"
+	clitestutil "github.com/cosmos/cosmos-sdk/testutil/cli"
```
aljo242, line 6: Import location

aljo242 (commented):
Hi @gjermundgaraba !

Been following the chain development for a while and wanted to check out how you were going about the sdk upgrade.  Just made some tiny suggestions

chain/Makefile
gjermundgaraba: Awesome, fixed that! :)

aljo242 (approved):

maheshKRex (approved):
lgtm

ba1ciu (approved):

PR #446, opened 2023-01-24
e2e tests for issue credits and collection balance transfer

maheshKRex:
drafting to check if the local issue with getCLIResponse shows up at server.


gjermundgaraba:
@maheshKRex it looks like a merge has gone bad here. Your branch is undoing a bunch of things, including the upgrade to 0.47. That might also be why you are having trouble with the getCLIResponse?

maheshKRex:
> @maheshKRex it looks like a merge has gone bad here. Your branch is undoing a bunch of things, including the upgrade to 0.47. That might also be why you are having trouble with the getCLIResponse?

yes, i made a mistake during raising the PR, closed it and created new PR https://github.com/EmpowerPlastic/empowerchain/pull/447

PR #461, opened 2023-01-30
455 add back ibc

gjermundgaraba:


chain/scripts/cosmovisor_upgrade_serve.sh
```diff
@@ -52,7 +52,7 @@ mkdir -p $TMP_COSMOVISOR_DIR/genesis/bin
 mkdir -p $TMP_COSMOVISOR_DIR/upgrades/v2/bin
 cp $V1_BINARY $TMP_COSMOVISOR_DIR/genesis/bin/
 
-make build
+make -B build
```
ba1ciu, line 55: Just curious, why is it needed?

ba1ciu (approved):

chain/scripts/cosmovisor_upgrade_serve.sh
gjermundgaraba: I'm not 100% sure that it is. I added it during some debugging, and it doesn't hurt. Basically it makes sure to bypass make's "caching" mechanism. Perhaps this would never actually happen, but idunno.

gjermundgaraba: I looked more into it, and we actually have a mechanism already to avoid this problem with ".PHONY: *" in the makefile. So removed all of these -B uses

PR #513, opened 2023-02-23
477 clean up html css

PiranavanShanmugavadivelu:
Initiated the Tailwind CSS and basic routing


package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
gjermundgaraba, line 1: This one should probably not be here. I think you can delete it?

frontend/proof-of-existence/.husky/pre-commit
```diff
@@ -1,5 +0,0 @@
-#!/usr/bin/env sh
```
gjermundgaraba, line 1: Maybe we can keep the .husky setup to make sure we get lint on pre-commits?

frontend/proof-of-existence/package-lock.json
```diff
@@ -0,0 +1,5284 @@
+{
```
gjermundgaraba, line 1: Since we are using Lerna with workspaces, there should not be any package-lock file here. Ask me if you have any questions about this process.

frontend/proof-of-existence/package.json
```diff
@@ -3,7 +3,6 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "prepare": "cd ../.. && husky install frontend/proof-of-existence/.husky",
```
gjermundgaraba, line 6: If we keep husky, we should keep this one too.

frontend/proof-of-existence/src/assets/logo.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 261.76 226.69"  xmlns:v="https://vecta.io/nano"><path d="M161.096.001l-30.225 52.351L100.647.001H-.005l130.877 226.688L261.749.001z" fill="#41b883"/><path d="M161.096.001l-30.225 52.351L100.647.001H52.346l78.526 136.01L209.398.001z" fill="#34495e"/></svg>
```
gjermundgaraba, line 1: This file should probably be deleted.

gjermundgaraba (changes requested):
Code looks good, just a few small things to delete and add back. See the comments.

gjermundgaraba (approved):

PR #526, opened 2023-03-01
prepare-genesis cmd

ba1ciu:
New binary subcommand `prepare-genesis` that takes a `genesis.json` file as input and modifies it accordingly to the preset. For now the only available preset is devnet and in that case genesis staking state is reset to default, genutil transaction for devnet is added and E2E Plastic Credit Test data is added.

The idea for genesis tool was to create a struct of all modules' genesis states, so we can have different functions that can modify subsets of genesis states. There's one flow to this solution as genesis transformations can overwrite each other if not carefully implemented. Open to suggestions.

I had to rename currently used EMP and PCRD Credit Types in E2E, so when we'll run the tool on an actual mainnet genesis, we won't have a clash of abbreviations.


ba1ciu:
Closing for now. Using genutil for devnet preset is hardly usable for anyone other than us, who're owning the devnet validator private key. I'll change it to using stake.validators and will provide an interactive way to put your own tendermint/cosmos keys.

ba1ciu:
Now after the update, user can provide a key name from keyring, which will become a validator operator key and a consensus public key from priv_validator_key.json as cli args.

chain/cmd/empowerd/cmd/genesis.go
```diff
@@ -0,0 +1,97 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	genutiltypes "github.com/cosmos/cosmos-sdk/x/genutil/types"
+)
+
+// PrepareGenesisCmd returns prepare-genesis cobra Command.
+
+func PrepareGenesisCmd(defaultNodeHome string, mbm module.BasicManager) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "prepare-genesis",
```
gjermundgaraba, line 25: `Use` should have the params in it

gjermundgaraba (changes requested):
Why does this require an existing genesis file?

I was actually thinking: what if we could avoid having a manual genesis file at all? Sure, in the end there needs to be one, but instead of having on that is manually edited and so on, it could be done entirely in code. `prepare-genesis` could simply spit out a ready-to-use genesis file.

gjermundgaraba:
> I had to rename currently used EMP and PCRD Credit Types in E2E, so when we'll run the tool on an actual mainnet genesis, we won't have a clash of abbreviations.

This I also didn't understand. An actual mainnet genesis (at least the initial one) would not have any credit types yet, right?

ba1ciu:
@gjermundgaraba I added subcommand for mainnet generation from scratch and there's a placeholder function for tweaking module params. Should we implement those tweaks now or as a separate PR?

gjermundgaraba:
> @gjermundgaraba I added subcommand for mainnet generation from scratch and there's a [placeholder function](https://github.com/EmpowerPlastic/empowerchain/pull/526/files#diff-13ce7bda66a9e813daef33c69d77a598623c8423a78978e65819dfef59893320) for tweaking module params. Should we implement those tweaks now or as a separate PR?

Up to you!

ba1ciu:
Let's add specific mainnet params in the separate PR and review the current state of genesis tool now.

gjermundgaraba:
@ba1ciu should the "genesis-tools" directory instead just be a "genesis" directory under /app? I am not sure if it makes sense to have it as another top-level directory?

ba1ciu:
> @ba1ciu should the "genesis-tools" directory instead just be a "genesis" directory under /app? I am not sure if it makes sense to have it as another top-level directory?

Yes, makes sense to put it under /app. Will do that.

gjermundgaraba (changes requested):
I've left a couple of questions.

Also, I would like to see some kind of regression test for this so that if we mess something up it would get caught (invalid genesis outputs, for instance).

ba1ciu:
Moved subcommands under `genesis` command. Added e2e tests. Added utils for simpler account/bank genesis state modification.

gjermundgaraba (approved):

gjermundgaraba:
@ba1ciu wat dis? I assume it's a mistake, let's remove it

image

gjermundgaraba (changes requested):
Remove the ignite commit

ba1ciu:
> @ba1ciu wat dis? I assume it's a mistake, let's remove it
> 
> 
> 
> image
> 
> 

Actually I was generating some Ignite helpers and forgot that it makes a commit every time you generate something with it. So it committed my changes in it's name. I'll try to revert the commit and add those changes as mine.

gjermundgaraba:
Ok makes sense. You can rebase them locally probably. Feel free to squash them as it is what I am going to do when mergin it in anyways

gjermundgaraba (approved):

PR #530, opened 2023-03-06
IDCounters invariant

ba1ciu:
Turned out that invariant is being checked before state is being updated if running fresh, so I had to exclude the case when the IDCounters cannot be fetched from the store altogether. Open to suggestions to deal with that other way.


ba1ciu:
Turned out the issue is only with unit tests. Disabled genesis invariant checking for them to work properly as we don't need it anyway.

ba1ciu:
At the end, it turned out we have to change the order of genesis initialization, so crisis is being initialized after plasticcredit as part of crisis' initializaion is verifying the invariants.

gjermundgaraba (approved):

PR #549, opened 2023-03-15
Update listing

ba1ciu:


chain/tests/e2e/marketplace/update_listing_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	wasmcli "github.com/CosmWasm/wasmd/x/wasm/client/cli"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	clitestutil "github.com/cosmos/cosmos-sdk/testutil/cli"
+
+	"github.com/EmpowerPlastic/empowerchain/tests/e2e"
+	"github.com/EmpowerPlastic/empowerchain/x/plasticcredit/client/cli"
+)
+
+func (s *E2ETestSuite) TestUpdateListing() {
```
gjermundgaraba, line 14: I prefer to keep the e2e tests lighter and only test the things that can't be unit-tested directly in rust. These tests are very slow to run, after all.

Many of these tests seem redundant, given that they also exist in the unit tests, right?

cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
```diff
@@ -28,17 +29,18 @@ pub fn execute_create_listing(
         return Err(ContractError::ZeroPrice {});
     }
 
-    let next_listing_id = NEXT_LISTING_ID.load(deps.storage)?;
+    if (LISTINGS.may_load(deps.storage, (info.sender.clone(), denom.clone()))?).is_some() {
```
gjermundgaraba, line 32: Can you add a unit test that covers this scenario?

cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
```diff
@@ -100,14 +102,78 @@ pub fn execute_buy_credits(
...
+    }
+
+    // Update the listing
+    listing.number_of_credits = number_of_credits;
+    listing.price_per_credit = price_per_credit.clone();
+    LISTINGS.save(deps.storage, (listing.owner.clone(), listing.denom.clone()), &listing)?;
+
+    let res = Response::new()
+        .add_attribute("action", "update_listing")
+        .add_attribute("listing_owner", info.sender)
```
gjermundgaraba, line 166: Probably denom would be a good idea to add here too?

cosmwasm/contracts/plastic-credit-marketplace/src/query.rs
```diff
@@ -8,29 +8,40 @@ pub const DEFAULT_LIMIT: u64 = 30;
 #[entry_point]
 pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
     match msg {
+        QueryMsg::Listing {owner, denom} => to_binary(&listing(deps, owner, denom)?),
```
gjermundgaraba, line 11: Maybe at least a simple unit test for this one too (not found + found)

gjermundgaraba (changes requested):

chain/tests/e2e/marketplace/update_listing_test.go
ba1ciu: > I prefer to keep the e2e tests lighter and only test the things that can't be unit-tested directly in rust. These tests are very slow to run, after all.
> 
> Many of these tests seem redundant, given that they also exist in the unit tests, right?

Agreed. In that case I'd only keep 2 first tests as there are actual plastic credit transfers being made there. Makes sense? @gjermundgaraba

gjermundgaraba: Yes, that makes perfect sense :+1:

gjermundgaraba: Off-topic for this PR, but maybe we should do this for all our other e2e tests too. See which ones are actually redundant. They are getting _veeeery_ slow and heavy now.

ba1ciu: Agreed! Let's follow it [here](https://github.com/EmpowerPlastic/empowerchain/issues/559)!

gjermundgaraba (approved):

Issue #559, opened 2023-03-17
Lighten e2e tests


ba1ciu:
Review e2e tests and delete redundant ones because of existing unit/integration tests.

maheshKRex:
Copying the nice summarisation of testing at different levels on chain, by @ba1ciu :
So we have:
1. Types and Messages tests (aka unit tests, types_test.go, msgs_test.go) - lowest level ones that should handle testing invalid message and object formats. Any higher level tests doing just that should be removed.
2. Genesis tests (genesis_test.go in module dir for testing object formats and genesis_test.go in keeper for testing init/export) - those are fine as they are and kinda separate from other tests.
3. Integration tests (msg_server_test.go and query_server_test.go) - middle level of tests (kinda controller level testing). Should use this for everything other than object/message validation. We can test interaction between different states in the same module thanks to PopulateWithSamples() . Good example of the test that should be done at this level.
4. Simulation tests - again, separate, should leave as it is.
5. Smoke tests - also separate. I'd vote for removing them entirely as they are doing the same as E2E tests, but wait for Gjermund with that.
6. E2E tests - highest level. Should be used when 1st and 3rd cannot be used.
So we should first test everything we can using 1, then other stuff using 3 and if there's still something to cover - 6.

PR #574, opened 2023-03-24
cleaned up e2e tests

maheshKRex:
https://github.com/EmpowerPlastic/empowerchain/issues/559


chain/tests/e2e/plasticcredit/applicant_tx_test.go
```diff
@@ -1,206 +0,0 @@
...
-			updatedTitle:      "Plastix Inc.",
-			updatedDesc:       "Grab that bottle",
-			fromFlagValue:     fmt.Sprintf("--%s=%s", flags.FlagFrom, issuerKey.Name),
-			expectedErrOnSend: false,
-			expectedErrOnExec: true,
-			expectedErrMsg:    "applicant not found",
-			expectedState:     nil,
-		},
-
-		"wrong signer": {
```
ba1ciu, line 143: We should add an integration test with a wrong signer, but one that will pass bech32 validation.

chain/tests/e2e/plasticcredit/issuer_tx_test.go
```diff
@@ -10,256 +10,13 @@ import (
...
-	"github.com/cosmos/gogoproto/proto"
 
 	"github.com/EmpowerPlastic/empowerchain/app/params"
 	"github.com/EmpowerPlastic/empowerchain/tests/e2e"
-	"github.com/EmpowerPlastic/empowerchain/testutil/sample"
 	"github.com/EmpowerPlastic/empowerchain/x/plasticcredit"
 	"github.com/EmpowerPlastic/empowerchain/x/plasticcredit/client/cli"
 )
 
-func (s *E2ETestSuite) TestCmdCreateIssuer() {
```
ba1ciu, line 22: I'd leave issuer creation with gov. We don't have that covered on the integration level.

chain/x/plasticcredit/keeper/msg_server_test.go
```diff
@@ -313,37 +303,41 @@ func (s *TestSuite) TestUpdateIssuer() {
 	}
 }
 
-func (s *TestSuite) TestCreateApplicant() {
+func (s *TestSuite) TestUpdateIssuerCreator() {
```
ba1ciu, line 306: It's the same as `TestCreateIssuer`?

ba1ciu (changes requested):

chain/tests/e2e/plasticcredit/applicant_tx_test.go
maheshKRex: yes, having this in msg_server_tests, here sample.AccAddress generates some valid address, but not authorised to sign.
`"unauthorized caller / wrong signer": {
			msg: &plasticcredit.MsgUpdateApplicant{
				Updater:     sample.AccAddress(), 
				ApplicantId: 1,
				Name:        "Empower",
				Description: "Empower is cool",
				Admin:       issuerAdmin,
			},
			err: sdkerrors.Er`

chain/tests/e2e/plasticcredit/issuer_tx_test.go
maheshKRex: agree. reverted the removal.

chain/x/plasticcredit/keeper/msg_server_test.go
maheshKRex: nice catch. removed them.

chain/tests/e2e/plasticcredit/issuer_tx_test.go
```diff
@@ -10,13 +10,256 @@ import (
...
+				s.Require().NoError(val.ClientCtx.Codec.UnmarshalJSON(out.Bytes(), &resp))
+				s.Require().Equal(tc.expectedState.Name, resp.Issuers[len(resp.Issuers)-1].Name)
+				s.Require().Equal(tc.expectedState.Description, resp.Issuers[len(resp.Issuers)-1].Description)
+				s.Require().Equal(tc.expectedState.Admin, resp.Issuers[len(resp.Issuers)-1].Admin)
+			}
+		})
+	}
+}
+
+func (s *E2ETestSuite) TestCmdUpdateIssuer() {
```
ba1ciu: Updates we can get rid of I think as those are done by sending a standard message and leave only creates as they use `NewCmdSubmitProposal()`?

ba1ciu (changes requested):

ba1ciu (approved):

PR #581, opened 2023-04-03
Upgrade to 47 + cometbft + wasmd RC

gjermundgaraba:


chain/app/simulation/app_state_determinism_test.go
```diff
@@ -48,13 +50,18 @@ func TestAppStateDeterminism(t *testing.T) {
 			}
 
 			db := dbm.NewMemDB()
-			empowerApp := app.New(logger, db, nil, true, map[int64]bool{},
-				app.DefaultNodeHome,
-				simcli.FlagPeriodValue,
-				params.MakeEncodingConfig(app.ModuleBasics),
-				simtestutil.EmptyAppOptions{},
+			appOptions := make(simtestutil.AppOptionsMap, 0)
+			appOptions[flags.FlagHome] = t.TempDir() // ensure a unique folder
```
ba1ciu, line 54: So this was it? Tests were reusing the same state?

ba1ciu (approved):

chain/app/simulation/app_state_determinism_test.go
gjermundgaraba: I believe so

gjermundgaraba: We'll see :see_no_evil:

Issue #556, opened 2023-03-16
Cancel listing message


gjermundgaraba:
Add a new message to the marketplace contract that allows for cancel of an existing listing.

### Cancel listing message:
```rust
CancelListing {
   denom: String,
}
```
### Action response:
- action: cancel_listing
- cancelled_by: <owner>
- denom: <denom>

### Logic:
1. Fetch listing by <owner, denom>.
2. ListingNotFound, if it's already cancelled.
3. Transfer the balance from contract to listing owner.
4. Cancel the listing.

### Other changes:
1. At buy credits, cancel the listing when all the credits in the listing are bought.

PR #587, opened 2023-04-05
added cancel listing and tests

maheshKRex:
implementation for [cancel listing](https://github.com/EmpowerPlastic/empowerchain/issues/556)

### DIFF
--- a/chain/tests/e2e/marketplace/cancel_listing_test.go
+++ b/chain/tests/e2e/marketplace/cancel_listing_test.go
@@ -1,0 +1,86 @@
+package marketplace_test
+
+import (
+	"fmt"
+
+	wasmcli "github.com/CosmWasm/wasmd/x/wasm/client/cli"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	clitestutil "github.com/cosmos/cosmos-sdk/testutil/cli"
+
+	"github.com/EmpowerPlastic/empowerchain/tests/e2e"
+	"github.com/EmpowerPlastic/empowerchain/x/plasticcredit/client/cli"
+)
+
+func (s *E2ETestSuite) TestCancelListing() {
+	val := s.Network.Validators[0]
+	// Instantiate marketplace contract
+	marketplaceAddress := s.instantiateMarketplace()
+	creditOwnerKey, err := val.ClientCtx.Keyring.Key(e2e.ApplicantKeyName)
+	s.Require().NoError(err)
+	creditOwnerAddress, err := creditOwnerKey.GetAddress()
+	s.Require().NoError(err)
+
+	testCases := map[string]struct {
+		denom                        string
+		expectedCreatorCreditBalance uint64
+		expectedErr                  bool
+	}{
+		"happy path": {
+			denom:                        "PTEST/UPDATE_LISTING_1",
+			expectedCreatorCreditBalance: 1000,
+			expectedErr:                  false,
+		},
+	}
+	for name, tc := range testCases {
+		s.Run(name, func() {
+			// Grant transfer permissions to the marketplace contract
+			grantTransferCmd := cli.MsgGrantTransferAuthorizationCmd()
+			out, err := clitestutil.ExecTestCLICmd(val.ClientCtx, grantTransferCmd, append([]string{
+				marketplaceAddress,
+				tc.denom,
+				"100000",
+				fmt.Sprintf("--%s=%s", flags.FlagFrom, creditOwnerKey.Name),
+			}, s.CommonFlags...))
+			s.Require().NoError(err, out.String())
+			cliResponse, err := s.GetCliResponse(val.ClientCtx, out.Bytes())
+			s.Require().NoError(err)
+			s.Require().Equal(uint32(0), cliResponse.Code, cliResponse.RawLog)
+
+			// Create the listing
+			executeContractCmd := wasmcli.ExecuteContractCmd()
+			out, err = clitestutil.ExecTestCLICmd(val.ClientCtx, executeContractCmd, append([]string{
+				marketplaceAddress,
+				fmt.Sprintf(`{"create_listing": {"denom": "%s", "number_of_credits": "100", "price_per_credit": {"denom": "umpwr", "amount": "10"}}}`, tc.denom),
+				fmt.Sprintf("--%s=%s", flags.FlagFrom, creditOwnerKey.Name),
+			}, s.CommonFlags...))
+			s.Require().NoError(err, out.String())
+			cliResponse, err = s.GetCliResponse(val.ClientCtx, out.Bytes())
+			s.Require().NoError(err)
+			s.Require().Equal(uint32(0), cliResponse.Code, cliResponse.RawLog)
+
+			// Cancel the listing
+			out, err = clitestutil.ExecTestCLICmd(val.ClientCtx, executeContractCmd, append([]string{
+				marketplaceAddress,
+				fmt.Sprintf(`{"cancel_listing": {"denom": "%s"}}`, tc.denom),
+				fmt.Sprintf("--%s=%s", flags.FlagFrom, creditOwnerKey.Name),
+			}, s.CommonFlags...))
+			s.Require().NoError(err, out.String())
+			cliResponse, err = s.GetCliResponse(val.ClientCtx, out.Bytes())
+			s.Require().NoError(err)
+			if tc.expectedErr {
+				s.Require().NotEqual(uint32(0), cliResponse.Code, cliResponse.RawLog)
+			} else {
+				s.Require().Equal(uint32(0), cliResponse.Code, cliResponse.RawLog)
+
+				// Check that the credit balances
+				creditOwnerBalanceAfter := s.GetCreditBalance(e2e.ApplicantAddress, tc.denom)
+				s.Require().Equal(tc.expectedCreatorCreditBalance, creditOwnerBalanceAfter.Active)
+
+				// Check the cancelled listing is removed
+				queryCmd := wasmcli.GetCmdGetContractState()
+				out, err = clitestutil.ExecTestCLICmd(val.ClientCtx, queryCmd, []string{"smart", marketplaceAddress, fmt.Sprintf(`{"listing": {"owner": "%s", "denom": "%s"}}`, creditOwnerAddress.String(), tc.denom)})
+				s.Require().Error(err, ErrListingNotFound)
+			}
+		})
+	}
+}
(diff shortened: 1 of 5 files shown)


cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
```diff
@@ -175,6 +179,36 @@ fn execute_update_listing(
...
 
+fn execute_cancel_listing(
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    denom: String,
+) -> Result<Response, ContractError> {
+    let listing = LISTINGS.load(deps.storage, (info.sender.clone(), denom.clone())).map_err(|_| ContractError::ListingNotFound {})?;
+
+    // Zero credits means the listing is already cancelled
```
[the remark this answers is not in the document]
gjermundgaraba: If I am wrong, please add a test that hits this path.

cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
```diff
@@ -868,5 +902,72 @@ mod tests {
...
+            let owner = creator_info.sender.clone();
+            execute(deps.as_mut(), mock_env(), creator_info.clone(), create_listing_msg).unwrap();
+
+            let cancel_listing_msg = ExecuteMsg::CancelListing {
+                denom: "pcrd".to_string(),
+            };
+            execute(deps.as_mut(), mock_env(), creator_info.clone(), cancel_listing_msg).unwrap();
+            
+            let listing = LISTINGS.load(deps.as_ref().storage, (owner, "pcrd".to_string())).map_err(|_| ContractError::ListingNotFound {});
+            assert_eq!(listing.unwrap_err(), ContractError::ListingNotFound {});
```
gjermundgaraba, line 940: Maybe also try to cancel the listing again and verify that you get ListingNotFound

cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
```diff
@@ -868,5 +902,72 @@ mod tests {
...
+                    amount: Uint128::from(1337u128),
+                },
+            };
+            let owner = creator_info.sender.clone();
+            execute(deps.as_mut(), mock_env(), creator_info.clone(), create_listing_msg).unwrap();
+
+            let cancel_listing_msg = ExecuteMsg::CancelListing {
+                denom: "pcrd".to_string(),
+            };
+            execute(deps.as_mut(), mock_env(), creator_info.clone(), cancel_listing_msg).unwrap();
```
gjermundgaraba, line 937: Check the attributes and messages sent by this (should be similar in other messages we have)

chain/tests/e2e/marketplace/cancel_listing_test.go
```diff
@@ -0,0 +1,86 @@
...
+			// Cancel the listing
+			out, err = clitestutil.ExecTestCLICmd(val.ClientCtx, executeContractCmd, append([]string{
+				marketplaceAddress,
+				fmt.Sprintf(`{"cancel_listing": {"denom": "%s"}}`, tc.denom),
+				fmt.Sprintf("--%s=%s", flags.FlagFrom, creditOwnerKey.Name),
+			}, s.CommonFlags...))
+			s.Require().NoError(err, out.String())
+			cliResponse, err = s.GetCliResponse(val.ClientCtx, out.Bytes())
+			s.Require().NoError(err)
+			if tc.expectedErr {
```
gjermundgaraba, line 70: And since we don't have any scenarios where we really need this, we can just remove the the `if` here :)

gjermundgaraba (changes requested):

chain/tests/e2e/marketplace/cancel_listing_test.go
```diff
@@ -0,0 +1,86 @@
...
+		expectedCreatorCreditBalance uint64
+		expectedErr                  bool
+	}{
+		"happy path": {
+			denom:                        "PTEST/UPDATE_LISTING_1",
+			expectedCreatorCreditBalance: 1000,
+			expectedErr:                  false,
+		},
+	}
+	for name, tc := range testCases {
```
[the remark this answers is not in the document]
maheshKRex: dropped the table, and also updated this to check owners credit balance, after cancel listing.

cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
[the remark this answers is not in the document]
maheshKRex: yes, added a test and when the listing is already cancelled it returns listing not found. So since the buy credits function cleans up any listing reaching a zero balance, and if there is no other user case to bring the balance to zero, this check may never get executed. should i remove it @gjermundgaraba

cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
maheshKRex: added test.

chain/tests/e2e/marketplace/cancel_listing_test.go
maheshKRex: removed.

cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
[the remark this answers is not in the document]
gjermundgaraba: Yeah, it seems like it should not be possible, so can remove it then :+1:

cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
```diff
@@ -100,6 +101,9 @@ pub fn execute_buy_credits(
             amount: total_price,
         }],
     });
+    if listing.number_of_credits.is_zero() {
+        LISTINGS.remove(deps.storage, (Addr::unchecked(owner.clone()), denom.clone()));
```
ba1ciu, line 105: If the credit balance reaches zero during the purchase, we have two storage operations, first to save the updated listing and then delete it with the next one. I assume it might not be very gas-efficient that way? Maybe let's use `if else` and perform update if the balance is gt 0 and delete if it's 0?

cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
```diff
@@ -868,5 +902,100 @@ mod tests {
...
+
+            let cancel_listing_msg = ExecuteMsg::CancelListing {
+                denom: "pcrd".to_string(),
+            };
+            let err = execute(deps.as_mut(), mock_env(), creator_info.clone(), cancel_listing_msg).unwrap_err();
+            assert_eq!(err, ContractError::ListingNotFound {});
+        }
+
+        #[test]
+        fn test_cancel_listing_by_non_owner() {
```
ba1ciu, line 957: This one looks incomplete. Also, there's no such thing as canceling by a non-owner, as we don't pass the owner address ourselves, so the test actually becomes same as `test_cancel_listing_that_does_not_exist`

chain/tests/e2e/marketplace/cancel_listing_test.go
```diff
@@ -0,0 +1,73 @@
...
+	creditOwnerKey, err := val.ClientCtx.Keyring.Key(e2e.ApplicantKeyName)
+	s.Require().NoError(err)
+	creditOwnerAddress, err := creditOwnerKey.GetAddress()
+	s.Require().NoError(err)
+
+	// Grant transfer permissions to the marketplace contract
+	grantTransferCmd := cli.MsgGrantTransferAuthorizationCmd()
+	out, err := clitestutil.ExecTestCLICmd(val.ClientCtx, grantTransferCmd, append([]string{
+		marketplaceAddress,
+		"PTEST/UPDATE_LISTING_1",
```
ba1ciu, line 27: Please use a separate credit collection for the test. It's gonna fail randomly if the execution order of this one and update listing changes.

chain/tests/e2e/marketplace/cancel_listing_test.go
```diff
@@ -0,0 +1,73 @@
...
+	out, err = clitestutil.ExecTestCLICmd(val.ClientCtx, executeContractCmd, append([]string{
+		marketplaceAddress,
+		fmt.Sprintf(`{"create_listing": {"denom": "%s", "number_of_credits": "100", "price_per_credit": {"denom": "umpwr", "amount": "10"}}}`, "PTEST/UPDATE_LISTING_1"),
+		fmt.Sprintf("--%s=%s", flags.FlagFrom, creditOwnerKey.Name),
+	}, s.CommonFlags...))
+	s.Require().NoError(err, out.String())
+	cliResponse, err = s.GetCliResponse(val.ClientCtx, out.Bytes())
+	s.Require().NoError(err)
+	s.Require().Equal(uint32(0), cliResponse.Code, cliResponse.RawLog)
+	creditOwnerBalanceBefore := s.GetCreditBalance(e2e.ApplicantAddress, "PTEST/00001")
```
ba1ciu, line 47: The balance before here is taken after the listing creation, so the user should have less credits, but it's compared to balance after, which is balance after the listing has been canceled. Doesn't seem right or am I mistaken?

ba1ciu (changes requested):

cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
maheshKRex: yeah, you got me using copilot too much, will remove it.

cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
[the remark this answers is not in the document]
maheshKRex: removed.

cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
maheshKRex: added checks.

cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
maheshKRex: yes, makes sense. added conditional.

chain/tests/e2e/marketplace/cancel_listing_test.go
maheshKRex: added a new collection.

chain/tests/e2e/marketplace/cancel_listing_test.go
maheshKRex: yes, was using the e2e.ApplicantAddress instead of using the owner's address , and i think that made it to return matching balance, not the deducted one. updated to capture balance before create listing and use correct address for both queries.

gjermundgaraba (approved):

PR #588, opened 2023-04-05
Local containerized empowerchain node

ba1ciu:


gjermundgaraba:
It might not be important, or even useful, but you could have done this in an Earthfile too - if you wanted to

.github/workflows/devnet_push.yml
```diff
@@ -0,0 +1,47 @@
...
+          version: 0.6.30
+
+      - name: Log in to registry
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u $ --password-stdin
+
+      - name: build image
+        run: earthly --ci --output +devnet
+        working-directory: ./chain
+
+      - name: push image
```
gjermundgaraba, line 32: You could do most of this in automatically in the earthfile when using `earhtly --push +devnet`. Check out the empower-platform Earthfiles to see some examples.

gjermundgaraba: https://github.com/EmpowerPlastic/empower-platform/blob/6181e9c1123d28e7dfe6864334a61ad088cd2900/.github/workflows/pipeline-be-user.yml#L106

gjermundgaraba: And you can have multiple tags in a single save image: https://github.com/EmpowerPlastic/empower-platform/blob/6181e9c1123d28e7dfe6864334a61ad088cd2900/general/backend/cloud-run-user/Earthfile#L21

gjermundgaraba: (Even if you add --push inside the Earthfile, it will not try to push unless you also add --push to the earthly command)

gjermundgaraba (changes requested):

.github/workflows/devnet_push.yml
ba1ciu: Cool! Will use that

ba1ciu:
Temporarily closing PR to not spam with actions

PR #592, opened 2023-04-10
478 input hash

PiranavanShanmugavadivelu:


ba1ciu (approved):
Lgtm, but let's wait for @gjermundgaraba and @jschill.

gjermundgaraba (approved):
Lookgs good to me, but like Jakub said, let's get @jschill to look over it as well, because my frontend eyes are very poor these days

jschill:
Can have a look tonight, im off grid atm

jschill (approved):
I think it looks good!

The only thing i have is the `redirectToWalletPage()` that expects a string as first argument (`hash`) but the where the function is invoked, its passing a possible nullish value (`redirectToWalletPage(result?.value, ...`). I think it might be better to throw an error if `result` is not set so we can catch it in our logs and the user wont be redirected to [the.url]/undefined?

gjermundgaraba:
@PiranavanShanmugavadivelu can you please look into the comment from Johannes, and as soon as that is sorted we'll get this merged :)

frontend/proof-of-existence/src/views/certify/CertifyView.vue
```diff
@@ -43,10 +44,28 @@ const readFile = async (file: File): Promise<ArrayBuffer> => {
...
   const result = window.empSha256(byteArray);
-  router.push({
-    path: `/proof/${result?.value}`,
-    query: { fileName: file.value?.name, time: new Date().getTime() },
-  });
+  redirectToWalletPage(result?.value, file.value?.name, new Date().getTime());
+};
+
+const redirectToWalletPage = (
+  hash: string,
```
jschill, line 51:
```suggestion
  hash?: string,
```

PR #594, opened 2023-04-10
Fix remaining credit class refs

gjermundgaraba:


maheshKRex (approved):
looks like i missed most of local variable renames in my run

ba1ciu (approved):

PR #600, opened 2023-04-11
589 fee variable on the marketplace

gjermundgaraba:


ba1ciu (changes requested):
Just found one place requiring formatting.

Also, I was thinking of the standard use case of royalty fee, where creator of the sold asset (issuer?) gets a cut from every transaction, which would of course require more complicated fee mechanism. Don't know if we would want that, so just leaving this for you to think about.

Issue #561, opened 2023-03-17
On-chain certificates


ba1ciu:
## Idea

Based on plastic credit retirement data defined in https://github.com/EmpowerPlastic/empowerchain/issues/523 we want to create a generic on-chain state for storing certificate data. The certificate doesn't exactly has to be related to plastic credit retirement.

## Design

Separate cosmos module called `certificates` with a state defined as follows:
```proto
message Certificate {
  uint64 id = 1;
  CertificateType type = 2;
  string owner = 3 [(cosmos_proto.scalar) = "cosmos.AddressString"];
  string issuer = 4 [(cosmos_proto.scalar) = "cosmos.AddressString"];
  map<string, string> data = 5;
}

enum CertificateType {
  CREDIT_RETIREMENT = 0;
} 

message MsgCreateCertificate {
  option (cosmos.msg.v1.signer) = "issuer";

  string issuer = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
  CertificateType type = 2;
  string owner = 3 [(cosmos_proto.scalar) = "cosmos.AddressString"];
  map<string, string> data = 4;
}

message QueryCertificateRequest {
  uint64 id = 1;
  string owner = 2 [(cosmos_proto.scalar) = "cosmos.AddressString"];
}

message QueryCertificateResponse {
  Certificate certificate = 1 [(gogoproto.nullable) = false];
}

message QueryAllCertificatesByUserRequest {
  string owner = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
  cosmos.base.query.v1beta1.PageRequest pagination = 2 [(gogoproto.nullable) = false];
}

message QueryAllCertificatesByUserResponse {
  repeated Certificate certificates = 1;
  cosmos.base.query.v1beta1.PageResponse pagination = 2 [(gogoproto.nullable) = false];
}
```

We should use certificate's owner address as a dynamic store key prefix to be able to fetch all certificates owned by a certain user.

Required keeper functions:
```go
func (k Keeper) CreateCertificate(ctx sdk.Context, certificate Certificate)
func (k Querier) Certificate(goCtx context.Context , req *QueryCertificateRequest)
func (k Querier) CertificatesByUser(goCtx context.Context , req *QueryAllCertificatesByUserRequest)
```

ba1ciu:
Initial idea for discussion FYI @gjermundgaraba

gjermundgaraba:
[...]

PR #624, opened 2023-04-20
561 on chain certificates

maheshKRex:
certificates as an on-chain module [task](https://github.com/EmpowerPlastic/empowerchain/issues/561)


chain/proto/empowerchain/certificates/query.proto
```diff
@@ -0,0 +1,41 @@
...
+  }
+}
+
+message QueryParamsRequest {}
+
+message QueryParamsResponse {
+  Params params = 1 [(gogoproto.nullable) = false];
+}
+
+message QueryCertificateRequest {
```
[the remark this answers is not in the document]
gjermundgaraba: I don't remember if we talked about queries for this module. If it is needed the query seems to be missing elsewhere too.

chain/x/certificates/client/cli/tx.go
```diff
@@ -0,0 +1,28 @@
...
+	"fmt"
+	"time"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/spf13/cobra"
+
+	"github.com/EmpowerPlastic/empowerchain/x/certificates"
+)
+
+var DefaultRelativePacketTimeoutTimestamp = uint64((time.Duration(10) * time.Minute).Nanoseconds())
```
gjermundgaraba, line 13: What is this for?

chain/x/certificates/client/cli/tx.go
```diff
@@ -0,0 +1,28 @@
...
+func GetTxCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                        certificates.ModuleName,
+		Short:                      fmt.Sprintf("%s transactions subcommands", certificates.ModuleName),
+		DisableFlagParsing:         true,
+		SuggestionsMinimumDistance: 2,
+		RunE:                       client.ValidateCmd,
+	}
+
+	// this line is used by starport scaffolding # 1
```
gjermundgaraba, line 25: Remove all of these comments

chain/x/certificates/genesis.go
```diff
@@ -0,0 +1,46 @@
...
+// failure.
+func (gs GenesisState) Validate() error {
+	mapCertificates := make(map[uint64]Certificate)
+	for _, certificate := range gs.Certificates {
+		if err := certificate.Validate(); err != nil {
+			return err
+		}
+
+		if _, exists := mapCertificates[certificate.Id]; exists {
+			return errors.Wrapf(ErrCertificateDuplicate, "duplicate issuer with id %d was found", certificate.Id)
```
gjermundgaraba, line 33: Not duplicate issuer, but duplicate certificate?

chain/x/certificates/genesis_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/EmpowerPlastic/empowerchain/testutil/sample"
+	"github.com/EmpowerPlastic/empowerchain/x/certificates"
+)
+
+func TestGenesisState_Validate(t *testing.T) {
+	validIssuer := sample.AccAddress()
+	testCases := []struct {
+		name string
+		gen  certificates.GenesisState
+		err  bool
```
gjermundgaraba, line 17: Maybe do the same as in plasticcredit genesis state test and check for the actual error (or nil). That, and instead of doing "name", do testCases := map[string]struct {

chain/x/certificates/keys.go
```diff
@@ -0,0 +1,55 @@
...
+	binary.LittleEndian.PutUint64(key, id)
+
+	return key
+}
+
+func CreateKeyFromString(key string) []byte {
+	return []byte(key)
+}
+
+func CreateCertificateKey(owner sdk.AccAddress, id uint64) ([]byte, error) {
```
gjermundgaraba, line 47: Instead of returning an error, just do what we do in plastic credit:
owner = address.MustLengthPrefix(owner)
That way you don't need to check the error everywhere

chain/x/certificates/module/module.go
```diff
@@ -0,0 +1,153 @@
...
+// RegisterServices registers a gRPC query service to respond to the module-specific gRPC queries
+func (am AppModule) RegisterServices(cfg module.Configurator) {
+	msgServer := keeper.NewMsgServerImpl(am.keeper)
+	certificates.RegisterMsgServer(cfg.MsgServer(), msgServer)
+
+	querier := keeper.Querier{Keeper: am.keeper}
+	certificates.RegisterQueryServer(cfg.QueryServer(), querier)
+}
+
+// RegisterInvariants registers the invariants of the module. If an invariant deviates from its predicted value, the InvariantRegistry triggers appropriate logic (most often the chain will be halted)
```
gjermundgaraba, line 120: delete

gjermundgaraba (changes requested):

chain/x/certificates/keeper/certificate.go
```diff
@@ -0,0 +1,123 @@
...
+	if len(bz) == 0 {
+		return certificates.Certificate{}, false
+	}
+	var certificate certificates.Certificate
+	k.cdc.MustUnmarshal(bz, &certificate)
+
+	return certificate, true
+}
+
+func (k Keeper) GetAllCertificatesByOwner(ctx sdk.Context, owner sdk.AccAddress, pageReq query.PageRequest) ([]certificates.Certificate, query.PageResponse, error) {
```
ba1ciu, line 32: `owner` is unused. We're iterating through every certificate in the store here. Not sure if it will work, but I'd try to fetch the certificate store by using a compound key of `<certificate_store, owner>`, then we should get results only for a given user. So we would to create a store key using function like [this one](https://github.com/EmpowerPlastic/empowerchain/blob/faba9f307b71f9fd7e23a7a1d4e6d2fe6f8e269b/chain/x/plasticcredit/keys.go#L47).

chain/x/certificates/keeper/msg_server.go
```diff
@@ -0,0 +1,54 @@
...
+func NewMsgServerImpl(keeper Keeper) certificates.MsgServer {
+	return &msgServer{Keeper: keeper}
+}
+
+func (m msgServer) UpdateParams(goCtx context.Context, req *certificates.MsgUpdateParams) (*certificates.MsgUpdateParamsResponse, error) {
+	if m.authority != req.Authority {
+		return nil, errors.Wrapf(sdkerrors.ErrUnauthorized, "invalid authority; expected %s, got %s", m.authority, req.Authority)
+	}
+	for _, issuer := range req.Params.AllowedIssuer {
+		if len(issuer) == 0 || sdk.AccAddress(issuer).Empty() {
```
ba1ciu, line 28: If we want to verify if it's a correct bech32 address, we should use `sdk.AccAddressFromBech32()`. `sdk.AccAddress()` expects public key and not bech32 address.

chain/x/certificates/keeper/msg_server_test.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+}
+
+func (s *TestSuite) TestCreateCertificate() {
+	testCases := map[string]struct {
+		msg *certificates.MsgCreateCertificate
+		err error
+	}{
+		"happy path": {
+			msg: &certificates.MsgCreateCertificate{
```
ba1ciu, line 81: I guess we might not want to allow `data` field to be empty. @gjermundgaraba ?

ba1ciu (changes requested):

chain/x/certificates/client/cli/tx.go
maheshKRex: removed

chain/x/certificates/genesis.go
maheshKRex: fixed.

chain/x/certificates/genesis_test.go
maheshKRex: updated.

chain/x/certificates/keys.go
maheshKRex: fixed.

chain/x/certificates/simulation/simap.go
```diff
@@ -0,0 +1,15 @@
+package simulation
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	simtypes "github.com/cosmos/cosmos-sdk/types/simulation"
+)
+
+// FindAccount find a specific address from an account list
+func FindAccount(accs []simtypes.Account, address string) (simtypes.Account, bool) {
```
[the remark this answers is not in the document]
maheshKRex: need to add simulation tests here, holding it until other priorities are fixed.

chain/x/certificates/client/cli/tx.go
maheshKRex: removed.

chain/x/certificates/module/module.go
maheshKRex: added invariant for IDCounter.

chain/app/app.go
maheshKRex, line 1: @gjermundgaraba @ba1ciu i'm getting an error on building the empowerd binary (the binary just report Error:  <nil>
on every command), on this branch, and it got something to do with the certificates module. 
- I basically added certificate module related entry in app.go, for every such module/keeper related entry done for plasticcredits.
- Or could it be an issue coming from /certificates/module.go, which is again i followed what we did for platiccredits.

chain/x/certificates/keeper/certificate.go
maheshKRex: fixed, by storing by owners in the KV store.

chain/app/app.go
maheshKRex: fixed with help of jakub, culprit was client/tx.go

chain/x/certificates/keeper/msg_server_test.go
gjermundgaraba: Yeah, makes sense

chain/x/certificates/keeper/msg_server.go
maheshKRex: updated.

chain/x/certificates/keeper/msg_server_test.go
maheshKRex: made it required field, updates validations.

chain/x/certificates/msgs.go
```diff
@@ -0,0 +1,55 @@
...
+	_, err = sdk.AccAddressFromBech32(m.Issuer)
+	if err != nil {
+		return errors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid issuer address (%s)", err)
+	}
+
+	return nil
+}
+
+func (m *MsgCreateCertificate) GetSigners() []sdk.AccAddress {
+	owner, err := sdk.AccAddressFromBech32(m.Owner)
```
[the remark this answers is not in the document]
maheshKRex: fixed.

chain/x/certificates/types.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	if _, err := sdk.AccAddressFromBech32(cert.Issuer); err != nil {
+		return errors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid certificate issuer address (%s)", err)
+	}
+
+	return nil
+}
+
+func (cert Certificate) AddressHasAuthorization(addr sdk.AccAddress) bool {
```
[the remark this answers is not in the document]
maheshKRex: removed.

chain/x/certificates/client/cli/tx.go
```diff
@@ -0,0 +1,65 @@
...
+		RunE:                       client.ValidateCmd,
+	}
+
+	cmd.AddCommand(MsgCreateCertificateCmd())
+	return cmd
+}
+
+func MsgCreateCertificateCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create-certificate [certificate-type] [certificate-owner] [certificate-issuer] [map of certificate attributes]",
```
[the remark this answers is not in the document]
maheshKRex: added.

chain/x/certificates/genesis.go
```diff
@@ -0,0 +1,46 @@
...
+		Certificates: []Certificate{},
+		IdCounters: IDCounters{
+			NextCertificateId: DefaultIndex,
+		},
+	}
+}
+
+// Validate performs basic genesis state validation returning an error upon any
+// failure.
+func (gs GenesisState) Validate() error {
```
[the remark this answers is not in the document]
gjermundgaraba: It looks like we do, right?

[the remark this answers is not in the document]
gjermundgaraba: Oh no, it says some of the error are never hit, maybe we need some more test cases to hit those too?

chain/x/certificates/keeper/certificate.go
```diff
@@ -0,0 +1,143 @@
...
+	params := k.GetParams(ctx)
+	if len(params.AllowedIssuer) > 0 {
+		if !slices.Contains(params.AllowedIssuer, req.Issuer) {
+			return 0, errors.Wrapf(sdkerrors.ErrUnauthorized, "invalid issuer %s", req.Issuer)
+		}
+	}
+	return k.CreateCertificateInternal(ctx, req)
+}
+
+func (k Keeper) CreateCertificateInternal(ctx sdk.Context, req *certificates.MsgCreateCertificate) (uint64, error) {
```
gjermundgaraba, line 89: why is the "internal" method public, while the other is private?

gjermundgaraba: In either case, maybe we can rename to better describe their difference?

gjermundgaraba (changes requested):

chain/x/certificates/client/cli/tx.go
```diff
@@ -28,38 +29,29 @@ func GetTxCmd() *cobra.Command {
 
 func MsgCreateCertificateCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "create-certificate [certificate-type] [certificate-owner] [certificate-issuer] [map of certificate attributes]",
+		Use:   "create-certificate [map of certificate attributes]",
```
ba1ciu, line 33: So are we reading certificate data from the file? If yes we should change cmd description and use absolute path

chain/scripts/test/testdata/certificate_update_params.json
```diff
@@ -4,7 +4,7 @@
       "@type": "/empowerchain.certificates.MsgUpdateParams",
       "authority": "empower10d07y265gmmuvt4z0w9aw880jnsr700jxwhwvd",
       "params": {
-        "allowed_issuers": [
+        "allowed_issuer": [
```
ba1ciu, line 7: Are we chainging this to just one issuer or why is this change made?

ba1ciu (changes requested):

chain/proto/empowerchain/certificates/query.proto
```diff
@@ -0,0 +1,62 @@
...
+
+option go_package = "github.com/EmpowerPlastic/empowerchain/x/certificates";
+
+service Query {
+  rpc Params(QueryParamsRequest) returns (QueryParamsResponse) {
+    option (google.api.http).get = "/EmpowerPlastic/empowerchain/certificates/params";
+  }
+
+  rpc Certificate(QueryCertificateRequest) returns (QueryCertificateResponse) {
+    option (google.api.http).get = "/EmpowerPlastic/empowerchain/certificates/{id}/{owner}";
```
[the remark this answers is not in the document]
ba1ciu: I agree. Pagination params should be in a request object and not query param.

chain/scripts/test/testdata/certificate_update_params.json
maheshKRex: looks like was mistaken set as "issuers" before, anyway now reverted back to "issuers" after renaming at types proto def.

chain/x/certificates/client/cli/tx.go
maheshKRex: updated to read from file

chain/x/certificates/module/module.go
```diff
@@ -0,0 +1,153 @@
...
+
+	keeper        keeper.Keeper
+	accountKeeper certificates.AccountKeeper
+	bankKeeper    certificates.BankKeeper
+}
+
+func NewAppModule(
+	cdc codec.Codec,
+	keeper keeper.Keeper,
+	ak certificates.AccountKeeper,
```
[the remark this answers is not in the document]
maheshKRex: removed.

chain/x/certificates/expected_keepers.go
```diff
@@ -0,0 +1,18 @@
+package certificates
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/auth/types"
+)
+
+// AccountKeeper defines the expected account keeper used for simulations (noalias)
+type AccountKeeper interface {
```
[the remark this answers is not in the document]
maheshKRex: removed.

chain/x/certificates/keeper/certificate.go
maheshKRex: when directly calling the keeper function createCertificate from plastic credits: retire credits, we discussed to skip the checks on allowed issuers, as the calling context of plastic credits does not have the allowed issuers params set. Just renamed this to CreateCertificateViaInternalModule, and it's going to be called from another module, so had to make it public. While the other function createCertificates with the allowed issuer check is used by certificates msg_server.

chain/x/certificates/genesis.go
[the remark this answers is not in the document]
maheshKRex: added

chain/x/certificates/client/cli/tx.go
```diff
@@ -0,0 +1,65 @@
...
+		Short: "create a new certificate",
+		Example: "empowerd tx certificates create-certificate  ./create_certificate.json\n" +
+			"create_certificate.json following format:\n" +
+			"{\"issuer\":\"<issuer_address>\",\"owner\":\"<owner_address>\",\"type\":\"<Certificate type>\",\"additional_data\":[{\"key\":\"test\",\"value\":\"test\"}]}",
+
+		Args: cobra.MinimumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			currentDir, err := filepath.Abs("./")
+			if err == nil {
+				fmt.Println("Error: ", err)
```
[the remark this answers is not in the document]
maheshKRex: updated

chain/proto/empowerchain/certificates/types.proto
```diff
@@ -0,0 +1,37 @@
...
+import "google/protobuf/timestamp.proto";
+import "cosmos/base/v1beta1/coin.proto";
+
+option go_package = "github.com/EmpowerPlastic/empowerchain/x/certificates";
+
+// Params defines the parameters for the module.
+message Params {
+  option (gogoproto.goproto_stringer) = false;
+
+  repeated string allowed_issuer = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
```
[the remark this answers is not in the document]
maheshKRex: yes. renamed

chain/proto/empowerchain/certificates/query.proto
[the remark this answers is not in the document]
maheshKRex: updated.

chain/x/certificates/keeper/certificate.go
gjermundgaraba: Hmm, maybe even better to have one that is called createCertificateFromAllowedIssuers and another just createCertificate. Or the other way around createCertificate (which checks allowed issuers) and one that says createCertificateSkipAllowedIssuers. Or something.
Got any good name ideas @maheshKRex  @ba1ciu

maheshKRex: updated.

gjermundgaraba (approved):

chain/x/certificates/msgs.go
```diff
@@ -0,0 +1,59 @@
...
+
+	return nil
+}
+
+func (m *MsgCreateCertificate) GetSigners() []sdk.AccAddress {
+	owner, err := sdk.AccAddressFromBech32(m.Issuer)
+	if err != nil {
+		panic(err)
+	}
+	return []sdk.AccAddress{owner}
```
ba1ciu, line 58:
```suggestion
func (m *MsgCreateCertificate) GetSigners() []sdk.AccAddress {
	issuer, err := sdk.AccAddressFromBech32(m.Issuer)
	if err != nil {
		panic(err)
	}
	return []sdk.AccAddress{issuer}
```
Nit-pick, I know 😛

chain/x/plasticcredit/keeper/credit.go
```diff
@@ -63,35 +65,64 @@ func (k Keeper) GetCreditBalances(ctx sdk.Context, pageReq query.PageRequest) ([
 	return creditBalances, *pageRes, nil
 }
 
-func (k Keeper) retireCreditsForAddress(ctx sdk.Context, owner sdk.AccAddress, denom string, amount uint64) (plasticcredit.CreditBalance, error) {
-	creditBalance, found := k.GetCreditBalance(ctx, owner, denom)
+func (k Keeper) retireCreditsForAddress(ctx sdk.Context, req *plasticcredit.MsgRetireCredits) (plasticcredit.CreditBalance, error) {
```
ba1ciu, line 68: Why did you pass the whole Msg object here? It's a bit inconsistent with the rest of the keepers where we "unpack" Msg objects at the msg server level and pass arguments directly to the keeper.

ba1ciu (changes requested):

chain/x/plasticcredit/keeper/credit.go
maheshKRex: after adding AdditionalData field to the certificate, we have 5 fields that need to be passed to keeper and save it there. So i thought why not pass the message itself as it has all the 5. wdyt? should we have each param passed @ba1ciu

ba1ciu: I get you and it's a lot of args, but I'd still pass each separately. Msg objects shouldn't reach into Keeper imo, but let us give the final vote to @gjermundgaraba.

gjermundgaraba: I have to agree with Jacob on this one. While I don't like all the args, the keeper should in "theory" not know about msgs. 5 arg function ftw :D

ba1ciu (approved):

PR #665, opened 2023-05-01
Added doc images + some more docs

gjermundgaraba:


gjermundgaraba:
@ba1ciu I will just merge this now because I need it deployed, but please do a review of the content in any case!

PR #667, opened 2023-05-03
First attempt with demo of empowerjs with marketplace

gjermundgaraba:


gjermundgaraba:
@ba1ciu let's look at this together when you are back. It works now, but the approach is a bit different.

ba1ciu (changes requested):
Looks 👌🏻 The only thing I would change is move `getEmpowerchainSigningClient` from proof-of-existence and into `empowerjs` and export it from there.

frontend/empowerjs-demo-application/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "private": true,
+  "scripts": {
+    "dev": "vite",
+    "build": "run-p type-check build-only",
+    "preview": "vite preview",
+    "build-only": "vite build",
+    "type-check": "vue-tsc --noEmit"
+  },
+  "dependencies": {
+    "@empower-plastic/empowerjs": "file:../empowerjs",
```
ba1ciu, line 13: Just wondering, if it's a demo app, maybe fetch from npm registry?

frontend/empowerjs/proto/cosmos/base/store/v1beta1/listening.proto
ba1ciu, line 1: Where this +x for all files come from?

ba1ciu (changes requested):
Just two questions :)

frontend/empowerjs-demo-application/package.json
gjermundgaraba: It didn't have the latest there I guess. How do we publish?

frontend/empowerjs/proto/cosmos/base/store/v1beta1/listening.proto
gjermundgaraba: Ah, I was running the generation script and it failed, if I remember correctly, because I didn't have access to delete a lot of the files.

frontend/empowerjs-demo-application/package.json
gjermundgaraba: I guess the problem I have is that the the marketplace stuff is in this branch - and it wont work until the stuff I have in this branch gets merged (and released into a new empowerjs)

I'll merge this, and then set the version.

PR #671, opened 2023-05-04
654 my credits page pc m 5

PiranavanShanmugavadivelu:


gjermundgaraba (approved):

frontend/marketplace/src/components/CreditCard.vue
```diff
@@ -0,0 +1,61 @@
...
+          <p class="text-title32 font-bold">100.00 ABCD</p>
+          <div class="grid grid-cols-2 gap-3">
+            <button
+                class="btn rounded-sm text-greenPrimary font-bold text-title14 bg-buttonBlack normal-case border-none">
+              View certificate
+            </button>
+            <button
+                class="btn rounded-sm text-greenPrimary font-bold text-title14 bg-buttonBlack normal-case border-none">
+              Retire credits
+            </button>
```
jschill, line 27: Two buttons with the same style calls for a <Button /> 😄

jschill (changes requested):

ba1ciu (changes requested):
I think we should already build it dynamically - not hardcode values into HTML, like hardcoded pagination in `MyCredits.vue` or the data on `CreditCard.vue`, but instead read it from variables and iterate if necessary.

frontend/marketplace/src/components/NavBar.vue
```diff
@@ -102,4 +102,10 @@
 
 </template>
 
+<style scoped>
+.nav-dropdown-button{
```
jschill, line 106: Nice!

frontend/marketplace/src/components/NavBar.vue
```diff
@@ -54,14 +54,14 @@
                 </div>
 
                 <div class="menu py-2 items-center w-full">
-                  <button
-                      class="btn btn-ghost rounded-sm text-greenPrimary font-bold text-title14 normal-case border-none">
+                  <Button
```
jschill, line 57: There is no Burton or is it registered globally?

PR #673, opened 2023-05-06
655 credit retirement screen pc m 6

PiranavanShanmugavadivelu:


gjermundgaraba (approved):

frontend/marketplace/src/components/RetireCreditsModal.vue
```diff
@@ -0,0 +1,67 @@
...
+
+        <div class="form-control">
+          <label class="label">
+            <span class="label-text text-white text-title18">Available credits in your account</span>
+          </label>
+          <label class="input-group rounded-sm w-fit border-dashed border-[1px] border-dashedBorderBlack">
+            <input type="text" placeholder="0.01" class="input text-white text-title24 bg-mediumGray w-[180px]" />
+            <span class="bg-mediumGray text-title16 text-dropDownText">$MPWR</span>
+          </label>
+        </div>
```
jschill, line 42: To avoid the repetition here, how about extracting this into a `<FormControl>` component where you can pass `label` and `placeholder` to it?

So the end would look like this:
```ts
<FormControl label="How many Plastic credits do you want to retire?" placeholder="0.01"/>
<FormControl label="Available credits in your account" placeholder="0.01"/>
```

jschill (changes requested):

ba1ciu (commented):
Other than @jschill's comments, lgtm!

frontend/marketplace/src/components/RetireCreditsModal.vue
PiranavanShanmugavadivelu: Some styles different for these two inputs so reused the styles without creating a new component

frontend/marketplace/src/components/RetireCreditsModal.vue
```diff
@@ -0,0 +1,84 @@
...
+
+      <div class="flex justify-between">
+        <p class="text-title38">
+          Retire credits
+        </p>
+      </div>
+
+      <div class="grid md:grid-cols-2 gap-5 mb-6 mt-2">
+        <div class="form-control">
+          <label class="label">
```
jschill, line 30: `for` attribute missing, since its not wrapping an input

frontend/marketplace/src/components/RetireCreditsModal.vue
```diff
@@ -0,0 +1,84 @@
...
+            <span class="input-label">How many Plastic credits do you want to retire?</span>
+          </label>
+          <label class="input-group input-box">
+            <input type="text" placeholder="0.01" class="input-text"/>
+            <span class="input-sub-text">$MPWR</span>
+          </label>
+        </div>
+
+        <div class="form-control">
+          <label class="label">
```
jschill, line 40: `for` attribute missing, since its not wrapping an input

jschill (changes requested):

frontend/marketplace/src/components/RetireCreditsModal.vue
jschill: If its not gonna be used somewhere else i guess its ok, but if its a component that is gonna be used in many places id vote for passing in the classes that differ as an object on a case-by-case basis. 

```ts

const mpwrInputCustomClass = {
  wrappingLabel: 'border-dashed border-[1px] border-dashedBorderBlack',
  standaloneLabel: '...',
  input: '...'
}

<FormControl
  label="Available credits in your account"
  placeholder="0.01"
  :customClass="mpwrInputCustomClass"
/>
```

I really liked the way we could group classes though, and maybe that could be abstracted into another layer, like a css/custom.css files or something so its globally reached - when needed.

gjermundgaraba (changes requested):
@PiranavanShanmugavadivelu some changes left to be fixed + conflict with main branch now, please resolve

PiranavanShanmugavadivelu:
> @PiranavanShanmugavadivelu some changes left to be fixed + conflict with main branch now, please resolve

Done

PR #674, opened 2023-05-08
Empower Data Format Builer (IPFS Index File)

ba1ciu:
Happy to hear any suggestions regarding the approach I took.


gjermundgaraba (approved):
I like it! Good use of the builder pattern!

PR #695, opened 2023-05-18
652 auction search results pc m 3

PiranavanShanmugavadivelu:
Auction page and custom components were created.


gjermundgaraba:
@ba1ciu, would it be possible to actually use real data here? If so I would like to start integrating data asap rather than just push frontend styles

PiranavanShanmugavadivelu:
> @ba1ciu, would it be possible to actually use real data here? If so I would like to start integrating data asap rather than just push frontend styles

Sure, We already started integrating GraphQL in a separate branch

gjermundgaraba:
Would you prefer to get this part merged in first, then get the GraphQL stuff in next?

PiranavanShanmugavadivelu:
> Would you prefer to get this part merged in first, then get the GraphQL stuff in next?

Ya, Created components with dummy data, So once real data is ready, we can pass that easily

ba1ciu (approved):

PR #706, opened 2023-05-23
minor change. use home path as a base dir

Northa:


testnets/circulus-1/README.md
```diff
@@ -11,6 +11,7 @@ PLEASE READ:
 
 1. **Install binary empowerd**
 ```bash
+cd $HOME
```
ba1ciu, line 14: It's not exactly necessary to clone repo to home. We don't aim to provide copy paste manual, but only a guidance, assuming that validators know they way around it and have their own solutions.

gjermundgaraba, line 14: We discussed these changes and decided that we didn't want to spell out or dictate to such an extent basic concepts which every validator should be able to do by themselves. If they aren't able to do this by themselves they are not experienced enough to be in gentx in any case - which is fine :)

So please revert the changes to the README file so we can merge the gentx :)

gjermundgaraba (changes requested):

ba1ciu (changes requested):

PR #719, opened 2023-05-25
Create gentx-bloclick.json

ivandiazperez:


gjermundgaraba:
@ivandiazperez I think the GH action failed here because you were using the main branch. Please redo this with the circulus-1 branch from the start :)

PR #773, opened 2023-05-26
Create Kalia-Network.json

KaliaNetwork:


gjermundgaraba:
You need to branch out from circulus-1 and target it back as well. Please redo :)

PR #800, opened 2023-06-01
Update validator-setup.md

Lefey:
Correct min-self-delegation and typo


ba1ciu (approved):
Good catch! Thanks @Lefey!

PR #830, opened 2023-06-07
Fix altrusitic-1/circulus-1 document references

Relyte:
This PR changes all instances of `altruistic-1` to `circulus-1` in the validator docs. Additionally, it adds a note regarding the cosmovisor path if installed while using `gvm` to manage go version installations. `gvm` tends to install go binaries into version specific directories rather than the typical `/home/$USER/go/bin` directory.

Edit: I've also pushed a new commit to clarify an issue when building `go1.20.3`. It is unable to be compiled with `go1.4` despite what gvm's documentation says. `go1.19` compiled without issue based on `go1.4`, so it's likely a change with `go1.20` specifically. To deal with this, I've recommended just downloading the binary version of `go1.19` and compiling `go1.20.3` using that since it doesn't result in this issue. I've also appended the `--default` flag to the `gvm use` command at the end so that the user doesn't need to remember to change their go profile on login for future updates.


ba1ciu (commented):
👍🏻

PR #832, opened 2023-06-07
Marketplace fixes 

PiranavanShanmugavadivelu:


gjermundgaraba:
@PiranavanShanmugavadivelu it looks like the build fails, can you take a look?

frontend/marketplace/.env
```diff
@@ -0,0 +1,5 @@
+VITE_API_ENDPOINT="https://devnet.empowerchain.io:3000/"
```
gjermundgaraba, line 1: This one should not be devnet. @ba1ciu you got the correct one?

frontend/marketplace/.env
```diff
@@ -0,0 +1,5 @@
...
+VITE_HTTPS_FILE_URL="https://devnet.empowerchain.io:8080/ipfs/"
```
gjermundgaraba, line 2: I assume this should also not point to devnet? @ba1ciu

gjermundgaraba (changes requested):

gjermundgaraba (approved):

ba1ciu (approved):

PR #862, opened 2023-06-10
From zero to hero validator guide

molla202:


gjermundgaraba:
I'm sorry, but this is a Turkish guide and it needs to be in English. It also would need to be more comprehensive, but please reach out on Discord and we can talk about what is needed.

Also, next time do not add gentx to the PR

PR #871, opened 2023-06-12
Create From zero to hero validator guide 

santosdios:


gjermundgaraba:
Hi @santosdios, would you be able to provide this as a markdown file? It's a bit easier to read and is the final format also.

PR #876, opened 2023-06-13
empowerd tx authz grant

AlexITProf:
when creating Grant authorization to my secondary wallet, a zero balance error occurs


gjermundgaraba:
It sounds like this should be an issue? Can you do it as an issue instead of a PR please? :)

https://github.com/EmpowerPlastic/empowerchain/issues/new

AlexITProf:
Done sir, thanks)

PR #907, opened 2023-06-18
893 certificate page

PiranavanShanmugavadivelu:


frontend/marketplace/src/components/NavBar.vue
```diff
@@ -77,6 +79,8 @@ const connect = async () => {
 </script>
 
 <template>
+<!--  To hide on certificate page-->
+  <template v-if="!router.meta?.hideNavFooter" >
```
ba1ciu, line 83: Just curious as I don't know Vue ways yet. Why additional `<template>` instead of adding `v-if` in the upper one?

ba1ciu (changes requested):
Just one question :)

frontend/marketplace/src/components/NavBar.vue
ba1ciu: Right! Got it!

ba1ciu (approved):

gjermundgaraba (approved):

PR #910, opened 2023-06-19
Create From zero to hero validator guide & Proof of existence

santosdios:


gjermundgaraba:
@santosdios these are getting really good! The only thing I would also like here is that the images themselves are part of the PR. I can always go back and download the pictures and add them myself, but if you have them anyway and can link to them directly instead of using a separate repo that would be great :)

santosdios:
accept
address
agregar
chain
error
faucet
Proof of exixtence guide_Mesa de trabajo 1_Mesa de trabajo 1
proof
saldo

santosdios:
buy
conect
conect2
credicts
retire
transfer
view

santosdios:
something like this? tell me if it's ok or do you need something else?

gjermundgaraba:
> something like this? tell me if it's ok or do you need something else?

Ideally the pictures would be next to the files and linked directly something like `Alt`

PR #912, opened 2023-06-19
refresh auction balance and my credit balance

ba1ciu:
I ended up using `refetch`, but there's a thing called `polling` in Apollo client, which would be a better fit for this. Unfortunatelly, I didn't manage to get it working with. If you want you can try using it @PiranavanShanmugavadivelu

https://www.apollographql.com/docs/react/data/queries/#polling

### DIFF
--- a/frontend/marketplace/src/components/CreditTabContent.vue
+++ b/frontend/marketplace/src/components/CreditTabContent.vue
@@ -87,9 +87,12 @@ const getCreditsData = async () => {
 
 const loadQueryData = (query: string) => {
   showSpinner.value = true
-  const {result, loading, error} = useQuery(gql`${query}`);
+  const {result, loading, error, refetch} = useQuery(gql`${query}`);
   data.value = {result, loading, error}
   showSpinner.value = false
+  setInterval(() => {
+    refetch()
+  }, 5000)
 }
 
 onMounted(() => {
--- a/frontend/marketplace/src/pages/AuctionDetails.vue
+++ b/frontend/marketplace/src/pages/AuctionDetails.vue
@@ -151,7 +151,6 @@ const getAuctionDetails = (id: string | string[]) => {
   showSpinner.value = false
   denom.value = result.value?.marketplaceListings?.nodes[0].denom
   owner.value = result.value?.marketplaceListings?.nodes[0].owner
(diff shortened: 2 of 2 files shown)


ba1ciu:
> @ba1ciu A small suggestion. What if we reload the data when the Retire credits successfully completed instead of using timeout? 🤔

It's not that simple, because we're not sure that indexed data has been already updated, there might be a delay between transaction passing successfully and indexer actually picking it up, so imo refetching periodically is a better solution.

PiranavanShanmugavadivelu:
> > @ba1ciu A small suggestion. What if we reload the data when the Retire credits successfully completed instead of using timeout? 🤔
> 
> It's not that simple, because we're not sure that indexed data has been already updated, there might be a delay between transaction passing successfully and indexer actually picking it up, so imo refetching periodically is a better solution.

@ba1ciu ,No, I mean from the front end side, Trigger reload in the retire credits success.

ba1ciu:
> > > @ba1ciu A small suggestion. What if we reload the data when the Retire credits successfully completed instead of using timeout? 🤔
> > 
> > 
> > It's not that simple, because we're not sure that indexed data has been already updated, there might be a delay between transaction passing successfully and indexer actually picking it up, so imo refetching periodically is a better solution.
> 
> @ba1ciu ,No, I mean from the front end side, Trigger reload in the retire credits success.

Yeah, I know :) It's just, if we reload immediately after we get a successful response from blockchain, there's no guarantee that it will return updated data, it can take some more time for query to take effect.

PiranavanShanmugavadivelu:
> > > > @ba1ciu A small suggestion. What if we reload the data when the Retire credits successfully completed instead of using timeout? 🤔
> > > 
> > > 
> > > It's not that simple, because we're not sure that indexed data has been already updated, there might be a delay between transaction passing successfully and indexer actually picking it up, so imo refetching periodically is a better solution.
> > 
> > 
> > @ba1ciu ,No, I mean from the front end side, Trigger reload in the retire credits success.
> 
> Yeah, I know :) It's just, if we reload immediately after we get a successful response from blockchain, there's no guarantee that it will return updated data, it can take some more time for query to take effect.

Oh ok, Then this is the better way 😄

PR #938, opened 2023-06-23
Create gentx-Nodeist.json

nodeiistt:


gjermundgaraba:
We are not accepting gentx yet

PR #979, opened 2023-06-29
gentx-Dexanode

Dexanode:


gjermundgaraba:
Hi, we are not accepting gentx as we are in mainnet

PR #1014, opened 2023-08-07
Changelog generation

ba1ciu:
I borrowed changelog generation and structure from (Stride)[https://github.com/Stride-Labs/stride/blob/main/.github/workflows/changelog.yml], but tweaked it a little so it fits into our repo.

The idea is:
- changelog generation action has to be run manually and old..new versions has to be provided, so diff can be generated
- then changelog can be manually tweaked a little if needed
- then we tag a commit and release is created (had to change release action to one supporting custom release description)


gjermundgaraba (approved):

gjermundgaraba:
Looks good, should we document the process more in the docs also?

ba1ciu:
@gjermundgaraba Added a section in docs. Please review it 🙏🏻

gjermundgaraba (approved):
Love it!

PR #1037, opened 2023-10-10
a

ademesin008:


gjermundgaraba:
EmpowerChain has been live on mainnet for a while, so gentx is not accepted at this time :)

PR #1072, opened 2023-11-25
chore(marketplace): Bump all packages and require node >= 20

jschill:
To avoid us falling behind, this PR will update all packages in package.json to its latest versions and also add a required engine of node >= 20.10.x (current latest LTS) and npm >= 10.2.3.


ba1ciu (approved):
🚀

PR #1077, opened 2023-11-28
5705 pay button

jschill:
This will
- ✅ move auth logic out of the individual vue files and into a composable to share its state (see stores/auth.ts)
- ✅ create a fetch helper so we can use native fetch instead of external lib
- ✅ start doing UX updates on the button
- ✅ start implementing the call to the api microservice for creating a transaction and generate a payment link
- ✅ move 3rd party lib toast into a helper


frontend/marketplace/src/components/BuyButton.vue
```diff
@@ -0,0 +1,101 @@
...
+  showButtonSpinner: boolean;
+  insufficientBalance: boolean;
+  coinFormatted: string;
+  handleCardPayment: () => void;
+  handleBuyCredits: () => void;
+  walletConnected: () => boolean;
+}
+const props = defineProps<BuyButtonProps>();
+
+const { isAuthenticated, handleSignIn } = useAuth();
```
PiranavanShanmugavadivelu, line 14: When the user is not logged in, `isAuthenticated ` returns an object.
```
{
    "_object": {
        "logtoClient": {
            "accessTokenMap": {},
            "logtoConfig": {
                "endpoint": "https://h3cwf8.logto.app/",
                "appId": "ilv0vuqy1u4zi05nyesgc",
                "prompt": "consent",
                "scopes": [
                    "openid",
                    "offline_access",
                    "profile",
                    "email"
                ],
                "resources": [
                    "http://localhost:3000"
                ]
            },
            "adapter": {
                "storage": {
                    "appId": "ilv0vuqy1u4zi05nyesgc"
                }
            }
        },
        "isAuthenticated": false,
        "loadingCount": 0,
        "error": {
            "name": "LogtoClientError",
            "code": "not_authenticated"
        }
    },
    "_key": "isAuthenticated",
    "__v_isRef": true
}
```   
So always showing `Pay with card` in the Buy button, May be any issue with Logto 🤔

jschill: @PiranavanShanmugavadivelu Great finding! 💯 👊  I will look into this right now, Thank you!

jschill: @PiranavanShanmugavadivelu Ah yes you are right! The `isAuthenticated` is not a boolean, its a `Ref<boolean>`, so when using it in the `<script>`-section of a vue component i need to use `isAuthenticated.value`. I think stricter linter rules would have cought this - but you are reviewing better than our linter!

When using a prefix like "is" and "has" (isAuthenticated, hasEmail, isCompleted....) I kind of expect those variables to be booleans - so i fooled myself this time! 🤡 🤡 🤡

PiranavanShanmugavadivelu (approved):

PiranavanShanmugavadivelu:
@jschill Build error 
image

ba1ciu (approved):

PR #1080, opened 2023-12-05
My certificates rework

jschill:
- ✅ Remove the tab switcher "My certificates/My credits" and leave only my certificates page available
- 🚧 Credits should be fetched using a different query if user is logged in using login/password method TBD @ba1ciu

Issue #5707


jschill:
This fix is in a new PR due to the conflicts. PR #1118

PR #1098, opened 2023-12-12
fix(marketplace): Update buy button when connecting to a wallet

jschill:
Move address to a global state so we can let the buy button react to it when it changes.

Fixes https://github.com/EmpowerPlastic/empower-platform/issues/5705#issuecomment-1845160437

### DIFF
--- a/frontend/marketplace/src/components/BuyCredits.vue
+++ b/frontend/marketplace/src/components/BuyCredits.vue
@@ -11,11 +11,12 @@ import {
   RPC_ENDPOINT,
 } from "@/config/config";
 import { onMounted, ref, watch, computed } from "vue";
-import { getWallet, walletConnected } from "@/utils/wallet-utils";
+import { getWallet } from "@/utils/wallet-utils";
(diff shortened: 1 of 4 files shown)


frontend/marketplace/src/components/BuyCredits.vue
```diff
@@ -39,6 +41,12 @@ const availableCreditsString = computed<string>(() => {
   return `${props.availableCredits}/${props.initialCredits}`;
 });
 
+watch(isWalletConnected, async (newVal) => {
```
PiranavanShanmugavadivelu, line 44: I think `isWalletConnected `is a ref value, So 
```
watch(isWalletConnected.value, async (newVal) => {
  if (newVal === true) {
    await initValues();
  }
});
```

jschill: Great finding! I was just about to change, then i did a quick google and it seems like watch() accepts a ref as the first argument.

PiranavanShanmugavadivelu (approved):

PR #1102, opened 2023-12-13
Run devnet deployment only manually

ba1ciu:


jschill (approved):
Wow really nice work!

PR #1105, opened 2023-12-14
fix(marketplace): Show materials details the same way on auction as i…

jschill:
…n auction list

This fixes the marketplace side of this issue: https://github.com/EmpowerPlastic/empower-platform/issues/5852

The last thing, with the material properties order in the platform can impossibly be important as of now?

### DIFF
--- a/frontend/marketplace/src/components/ProjectDetailMaterial.vue
+++ b/frontend/marketplace/src/components/ProjectDetailMaterial.vue
@@ -1,0 +1,50 @@
+<script setup lang="ts">
+import type { MaterialProperty } from "@/types/GraphqlSchema";
+import {
+  findPlasticTypeInMaterial,
+  stripPlasticTypeFromMaterial,
+  prettifyCardProperty,
+} from "@/utils/utils";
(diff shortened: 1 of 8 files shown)


frontend/marketplace/src/components/ProjectDetailMaterial.vue
```diff
@@ -0,0 +1,50 @@
...
+          class="text-title14"
+        >
+          {{ prettifyCardProperty(property) }}
+        </li>
+      </ul>
+    </div>
+  </div>
+</template>
+
+<!-- <div class="col-span-1 ...">
```
PiranavanShanmugavadivelu, line 33: Is there any reason to keep this🤔

PiranavanShanmugavadivelu (approved):

frontend/marketplace/src/components/ProjectDetailMaterial.vue
jschill: Absolutely not :-)
I will update asap

PR #1108, opened 2023-12-18
Marketplace operators

gjermundgaraba:


cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
```diff
@@ -221,6 +230,201 @@ fn execute_cancel_listing(
...
+    }
+
+    // We need the original_price to calculate the fee split
+    let original_price = listing.price_per_credit.amount.checked_mul(number_of_credits_to_release.into()).unwrap();
+    let original_price_as_coin = Coin {
+        denom: listing.price_per_credit.denom.clone(),
+        amount: original_price,
+    };
+    let (fee_split_msgs, _, fee_amount) = get_fee_split(deps.storage, original_price_as_coin).unwrap();
+    if fee_amount.amount > Uint128::from(0u128) {
```
ba1ciu, line 396: Does this mean we have to send the crypto payment with the message? Shouldn't we ignore any fee_split if there is one?

cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
```diff
@@ -221,6 +230,201 @@ fn execute_cancel_listing(
...
+        if info.funds.len() != 1 || info.funds[0].denom != listing.price_per_credit.denom || info.funds[0].amount < fee_amount.amount {
+            return Err(ContractError::NotEnoughFunds {});
+        }
+    }
+
+    if info.funds.len() > 0 && info.funds[0].amount > fee_amount.amount { // We can skip the denom check here because it is triggered in the previous if statement
+        return Err(ContractError::TooMuchFunds {});
+    }
+
+    let transfer_credits_msg = create_transfer_credits_from_contract_msg(
```
ba1ciu, line 406: We'll have to adjust this part as we decided that credits should be auto-retired for the buyer. Right now, it will just transfer the credits to the given account and we won't be able to retire them, no?

And then I guess we should also accept retirererName as a param for release_frozen_credits.

ba1ciu (changes requested):
Left some points for discussion

cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
gjermundgaraba: Not sure what you mean? The point here is to make sure the operator pay the fee (even though the rest of the payment is happening off-chain). The fee should still be paid. If there is not fee-split configured it will be zero of course.

cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
gjermundgaraba: That is a good point! Will add that.

cosmwasm/contracts/plastic-credit-marketplace/src/execute.rs
ba1ciu: Alright, I get you. It's just, my thinking was that if we're the operator, then we have to pay that fee for ourselves, right?

ba1ciu: I just wonder if that's the way we should go or maybe do something like a list of addresses not affected by fee_split? Dunno, loud thinking.

gjermundgaraba: Well, ideally the fee-split will be set up so that the chain gets a small cut of every purchase, in addition to us.

ba1ciu: Alright, so we indeed gonna need some usdc to pay the fees then.

ba1ciu (approved):

PR #1127, opened 2024-01-08
removed connect wallet request

PiranavanShanmugavadivelu:


jschill (changes requested):
If you remove the connect() function, we will never see if a user is authenticated via the wallet. Which means if you do a page refresh you will loose the wallet authenticated state? This will affect the buy button, you wont be able to buy using USDC$ or whatever its called on the blockchain.

jschill (approved):
👍

ba1ciu (approved):

PR #1138, opened 2024-01-14
Added price query + refactored price calculation

gjermundgaraba:


ba1ciu (approved):
Could you upgrade the current contract @gjermundgaraba ? :)

PR #1157, opened 2024-01-29
Update deps, fix auction detail data, fix google map when coordinates are wrong

jschill:
This will:
1. Update all deps in `package.json`
2. Make sure auction data is always refetched when going into an auction. This was not the case, it just fetched the data once and if you backed out from the page and go back it wouldnt update. This will hopefully fix a number of issues (such as: https://github.com/EmpowerPlastic/empower-platform/issues/6060 and https://github.com/EmpowerPlastic/empower-platform/issues/6059)
3. Get rid of the nasty Google maps error caused by us passing empty coordinates to vue3-google-maps. (`TypeError: Cannot read properties of null (reading 'fromLatLngToDivPixel')`)


frontend/marketplace/src/components/CustomGoogleMap.vue
```diff
@@ -14,6 +14,28 @@ const props = defineProps<CustomGoogleMapProps>();
...
+  if (
+    typeof obj !== "object" ||
+    obj === null ||
+    Object.keys(obj).length !== 2
+  ) {
+    return false;
+  }
+
+  // Check if the properties 'lat' and 'lng' exist and are not 0
+  if ("lat" in obj && "lng" in obj && obj.lat !== 0 && obj.lng !== 0) {
```
ba1ciu, line 28: But can't they be 0? Like when someone collects plastics roaming near the [Null Island](https://en.wikipedia.org/wiki/Null_Island)?

ba1ciu (changes requested):

frontend/marketplace/src/components/CustomGoogleMap.vue
jschill: LOL yea you might be right! 
We used two zeros as lat lng default value (`{ lat: 0, lng: 0 }`) in the state before we get real data via apollo. And that makes something to freak out. I have changed this now so default is null which is a more proper value for something that doesn't exist.

ba1ciu (approved):

PR #1158, opened 2024-01-29
Handle credit freeze events in indexer

ba1ciu:


indexer/src/mappings/mappingHandlers.ts
```diff
@@ -175,6 +178,139 @@ export async function handleCancelListing(event: CosmosEvent): Promise<void> {
...
+        buyer: buyer,
+        numberOfCreditsFrozen: numberOfCredits,
+        timeout: timeout,
+        timestamp: new Date(event.block.header.time.toISOString()),
+      });
+      await freezeCreditsWasmEvent.save();
+      const marketplaceListing = await MarketplaceListing.get(
+        `${listingOwner}-${denom}`
+      );
+      marketplaceListing.amount = marketplaceListing.amount - numberOfCredits;
```
jschill, line 305: Can this be decimal values?

jschill:
Where is the issue with the freeze credits?

indexer/src/mappings/mappingHandlers.ts
ba1ciu: Nope, just integers.

ba1ciu:
> Where is the issue with the freeze credits?

Wdym?

gjermundgaraba (approved):

jschill:
> > Where is the issue with the freeze credits?
> 
> Wdym?

Wonder where i can find the github issue so i can read up on what it is and the motivation for it.

ba1ciu:
> > > Where is the issue with the freeze credits?
> > 
> > 
> > Wdym?
> 
> Wonder where i can find the github issue so i can read up on what it is and the motivation for it.

I can't find the issue (looking for archive stuff is a real pain in github project..). We added an option to withdraw credits from marketplace listing without the need to pay for it in crypto for some authorized users called operators. It's so we can process fiat payment and withdraw credit and send it to the buyer. In the 1st version of the marketplace, you had to pay with crypto to get the credits, so we added the feature called freeze. It works like this:
1. User wants to buy credits with fiat.
2. We freeze the credit in listing so noone can buy it for some time.
3. We wait for the user to pay us with fiat.
4. When we register the payment, we release credits from the listing to user.

PR #1170, opened 2024-02-07
Wait longer for transaction to complete

ba1ciu:


ba1ciu:
@jschill Merging immediately as I'm sure it gives problems to anyone testing it right now.

PR #1172, opened 2024-02-08
Incorrect volume/image displayed for listing

alexander-astrand:


frontend/marketplace/src/components/AuctionResultsCard.vue
```diff
@@ -35,7 +36,11 @@ interface CardDetailsList {
 
 const props = defineProps<AuctionResultsCardProps>();
 const denom = ref("");
+const volume = computed(() => { 
+  return Number(props.cardData.creditCollection.retiredAmount) + Number(props.cardData.creditCollection.activeAmount)
+ })
```
jschill, line 41: Just to confirm. These amounts are never floating numbers, right?

Try this in your console: `Number(1.1)+Number(1.3)`

frontend/marketplace/src/components/AuctionCard.vue
```diff
@@ -26,15 +31,20 @@ const handleViewDetailsClick = () => {
...
-        convertIPFStoHTTPS(
-          auctionData?.creditCollection?.creditData?.nodes[0].mediaFiles
-            ?.nodes[0].url,
-        ) || auctionCard
-      "
-    />
+    <div v-if="showSpinner">
+      <CustomSpinner :visible="showSpinner" />
+    </div>
+    <div v-if="showSpinner === false">
```
jschill, line 37: `!showSpinner` to be consistent with the if above or `v-else` intead of `v-if`?

frontend/marketplace/src/utils/query-builder.ts
```diff
@@ -194,6 +194,8 @@ export class ListingsQueryBuilder {
                     pricePerCreditDenom
                     creditCollection{
                         creditType
+                        retiredAmount
+                        activeAmount
```
jschill, line 198: 👍

frontend/marketplace/src/components/ImageGallery.vue
```diff
@@ -2,32 +2,41 @@
...
   () => props.imageArray,
   (newValue) => {
     activeImageURL.value = newValue[0];
+    showSpinner.value = false;
   },
 );
 
 const handleActiveImage = (url: string) => {
   activeImageURL.value = url;
+  showSpinner.value = false;
```
jschill, line 25: Why is this needed?

jschill (approved):

frontend/marketplace/src/components/AuctionResultsCard.vue
alexander-astrand: I do get 2.4000000000000004 when doing that.. I see that parseInt is used in AuctionDetail.vue instead, is that a better solution?

frontend/marketplace/src/components/AuctionCard.vue
alexander-astrand: Fixed!

frontend/marketplace/src/components/ImageGallery.vue
alexander-astrand: It's not, fixed!

frontend/marketplace/src/components/AuctionResultsCard.vue
jschill: @alexander-astrand just wanted to point out that if you are working with decimals/floats in javascript you need to use Decimal.js (or similar library) or convert the value to an int (by multiplying) before you do the math

alexander-astrand: Thanks for pointing that out, I'll have a look at that! Do you reckon parseInt does the trick here, or would another solution be better? Not really any need for decimals here as I gathered it, and based on the component already using parseInt!

PR #1175, opened 2024-02-08
listing fee tooltip

PiranavanShanmugavadivelu:


ba1ciu (approved):

gjermundgaraba:
Not related to the PR, but so good that we can make these changes so quickly (from idea to done, even if it is pretty small change) 🚀

jschill (changes requested):
Awesome! But i think we should create a Tooltip-component so we don't accidently use different looks on the tooltips

frontend/marketplace/src/components/ui/Tooltip.vue
```diff
@@ -0,0 +1,19 @@
+<script setup lang="ts">
+interface Props {
+  label: string;
+  iconStyle: string;
+}
+const props = defineProps<Props>();
+</script>
+<template>
+  <p class="tooltip" :data-tip="props.label" data-tip-class="my-custom-tooltip">
```
jschill, line 9: Remove data-tip-class?

frontend/marketplace/src/components/ui/Tooltip.vue
```diff
@@ -0,0 +1,19 @@
+<script setup lang="ts">
+interface Props {
+  label: string;
+  iconStyle: string;
```
jschill, line 4: Maybe `iconClass` so we know its the class and not the style attribute that we're talking about

frontend/marketplace/src/components/ui/Tooltip.vue
```diff
@@ -0,0 +1,19 @@
...
+<template>
+  <p class="tooltip" :data-tip="props.label" data-tip-class="my-custom-tooltip">
+    <img :class="props.iconStyle" src="../../assets/questionIcon.svg" />
+  </p>
+</template>
+<style>
+.tooltip {
+  --tooltip-color: #000000;
+  --tooltip-text-color: #ffffff;
+  --tooltip-tail: 10px;
```
jschill, line 17: Niiiiiice

jschill (changes requested):

jschill (approved):

PR #1178, opened 2024-02-12
Tracking event type

ba1ciu:
Long time since we revisited plastic credit index file and we definitely should have info about transaction type in there.

### DIFF
--- a/frontend/empowerjs/src/plastic-credit-utils/empower-data-creator.ts
+++ b/frontend/empowerjs/src/plastic-credit-utils/empower-data-creator.ts
@@ -1,361 +1,359 @@
 import jsonschema from "jsonschema";
 import { CreditProp, EmpowerDataFormat } from "./empower-data-format";
 
-
 interface CreditPropBuilder {
-    build(): CreditProp;
-    reset(): void;
+  build(): CreditProp;
+  reset(): void;
 }
 
 export class PlasticCreditBuilder {
-    private creditEventsData: CreditProp;
-    private creditMediaData: CreditProp;
-    private creditFilesData: CreditProp;
-    private issuanceDate: Date;
-    private creditType: string;
-    private applicantData: CreditProp;
+  private creditEventsData: CreditProp;
+  private creditMediaData: CreditProp;
+  private creditFilesData: CreditProp;
+  private issuanceDate: Date;
+  private creditType: string;
+  private applicantData: CreditProp;
 
-    constructor() {
-        this.reset();
+  constructor() {
+    this.reset();
+  }
+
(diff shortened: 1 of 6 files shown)


indexer/schema.graphql
```diff
@@ -252,6 +252,7 @@ type EventData @entity {
   material: [MaterialData!]! @derivedFrom(field: "eventData")
   registrationDate: Date!
   creditData: CreditData!
+  type: String
```
jschill, line 255: This should be an enum?

frontend/empowerjs/src/plastic-credit-utils/empower-data-creator.ts
```diff
@@ -163,6 +164,11 @@ export class EventBuilder implements CreditPropBuilder {
         return this;
     }
 
+    setType(type: EventType): EventBuilder {
+        this.type = type;
+        return this;
+    }
```
jschill, line 170: We need to update the tests to use this function as well since `type` is required? `/empowerjs/test/empower-data.spec.ts`

frontend/empowerjs/src/plastic-credit-utils/empower-data-creator.ts
```diff
@@ -339,4 +353,9 @@ interface Property {
 interface File {
     name: string;
     url: string;
+}
+
+enum EventType {
+    RECEPTION = "reception",
+    DELIVERY = "delivery"
```
jschill, line 360: Where is the casing transformation happening? In the platform these are uppercased? Or are these type and `transactionType` two different things?

ba1ciu: These are different things. I didn't want to rely on transaction statuses from the platform, because they are gonna be changed anyway.

jschill: Makes sense. In that case i think we should have called it `TransactionEventType` or prefix the values `RECEPTION = "transaction_reception"` but now i assume an Event is something else than a Transaction.

indexer/schema.graphql
ba1ciu: I wasn't sure about this one, because of the inconsistency on the platform side, but I guess it makes more sense to go with Enum. Changed.

frontend/empowerjs/src/plastic-credit-utils/empower-data-creator.ts
jschill: Makes sense not to be too coupled to the transactions! 🎡

PR #1180, opened 2024-02-14
6008 seo and alignments

jschill:
NOTE! 
The design changes ive made in here will be thrown in the fire of death since my next branch will redesign the cards/listings anyway. So no need to do a review on those things, but please have a look at the other stuff, especially the named routes since those are essential for everything up next.
--

- Make sure we use `router-link` when using links so google can crawl
- Fix alignment issues and make sure the whole card is clickable

Alignment issue before:
Screenshot 2024-02-14 at 11 00 34

After:
Screenshot 2024-02-14 at 11 01 42


frontend/marketplace/src/utils/useListingVolume.ts
```diff
@@ -0,0 +1,29 @@
+import { computed, type Ref, toRefs, type ComputedRef } from "vue";
+import { type CreditCollection } from "@/types/GraphqlSchema";
+
+const isInteger = (value: number) => {
+  console.log("value", value, Math.round(value), Math.round(value) === value);
```
PiranavanShanmugavadivelu, line 5: I think we can remove the console.log()

PiranavanShanmugavadivelu (changes requested):
image

Colour changed in `View details ` button

jschill:
> image
> 
> Colour changed in `View details ` button

Button will be removed in another branch so im not fixing that

frontend/marketplace/src/router/index.ts
```diff
@@ -34,6 +34,21 @@ declare module "vue-router" {
...
 }
 
+// TODO: This is very similar to PageViewEvents - merge them
+export enum PageNames {
+  START_PAGE = "start page",
+  CERTIFICATES = "certificates",
+  FAQ = "faq",
+  LISTINGS = "listings",
+  LISTING = "listing",
+  LOGIN_CALLBACK = "login callback",
```
portlandnorway, line 44: Only concern is with the spaces in the 2 word route names. Seems unconventional, but if it works, then certainly not a stopper.

portlandnorway (approved):
I see no problem with the route stuff. Green light!

frontend/marketplace/src/router/index.ts
jschill: I agree with you on this. My OCD alarm is screaming. Need to solve it.

jschill: Ok i might have over complicated stuff now? But the routes are now snake cased. Please have a look @portlandnorway

jschill: @portlandnorway i cant live with these latest changes, im reverting and will fix this later (promise), but i simply dont really know how to do it without making a mess in typings. Maybe i'll just change the initial enum to kebab-/snake-case, but id like segment to use the same enum... so maybe a simple helper would be enough?

```ts
// snake case the enum
enum PageNames { 
  FAQ = "faq",
  LOGIN_CALLBACK = "login_callback"
  ...
}

// helpers/segment.ts
const toSegment = (pageName: PageNames): string => {
  return pageName.replaceAll('_', ' ')  
}


// used by router
router.push({ name: PageNames.LOGIN_CALLBACK })


// used by segment
analytics.trackPageView(toSegment(PageNames.LOGIN_CALLBACK))

jschill: Yeah i think the above would be enough, but i will add the `toSegment()` function inside the `analytics.trackPageView()` so we don't have to bother manually calling that all the time.

jschill: Thanks for the feedback Dennis and - LOL - making me spend 3 hours on code convention! 😄

jschill: (no i did not spend 3 hours on it)

portlandnorway: Haha this was fun to read, thanks Johannes! I like the conclusion you came to

PR #1184, opened 2024-02-14
Tracking event type

ba1ciu:


ba1ciu:
Pushing to main, because marketplace is down without it.

gjermundgaraba:
Looks good. Will it be the same one for testnet and mainnet (qa, prod)?

ba1ciu:
> Looks good. Will it be the same one for testnet and mainnet (qa, prod)?

Yeah. Stuff I managed to set up is I'm adding additional metadata based on environment to IPFS files, so if we ever want to clear only testnet files for example, that should be possible :)

PR #1185, opened 2024-02-15
fix(marketplace): Links to faq questions

jschill:
Now each faq question will gets its own link based on the id. This makes it possible for us to link to a specific question, scroll down to it and expand it.

### DIFF
--- a/frontend/marketplace/src/components/FAQCollapse.vue
+++ b/frontend/marketplace/src/components/FAQCollapse.vue
@@ -1,15 +1,62 @@
 <script setup lang="ts">
+import { onMounted, ref, computed } from "vue";
+import { useRoute, useRouter } from "vue-router";
 export interface FAQCollapseProps {
+  id: string;
   title: string;
   content: string;
+  name: string;
 }
-defineProps<FAQCollapseProps>();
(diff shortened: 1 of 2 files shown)


frontend/marketplace/src/components/FAQCollapse.vue
```diff
@@ -1,15 +1,62 @@
...
+        top: topPos, // scroll so that the element is at the top of the view
+        behavior: "smooth", // smooth scroll
+      });
+    }, 1);
+  }
+});
+
+const handleClick = () => {
+  if (isChecked.value) {
+    router.push({ hash: "" });
```
PiranavanShanmugavadivelu, line 35: I tried both `push` and `replace` locally, but both always scrolling to the top. Is there any solution to stop scrolling to top? 🤔

PiranavanShanmugavadivelu (changes requested):

frontend/marketplace/src/components/FAQCollapse.vue
jschill: Yeah its the router doing the scrolling after each route is completed. Fixed now by checking if the to and from link are the FAQpage.

PR #1198, opened 2024-02-27
6318 enable faq page

PiranavanShanmugavadivelu:
Issue - https://github.com/EmpowerPlastic/empower-platform/issues/6318


jschill (changes requested):
We need other FAQ IDs, that are based on the title (FAQ question) rather than an incremental number, since the order of the questions will change and then the numbers are wrong.

Ie "what_is_plastic_credits" rather than "faq_1" etc

ba1ciu (approved):

jschill (approved):
Awesome

PR #1200, opened 2024-02-27
Revert "image stretch issue fix"

ba1ciu:
Reverts EmpowerPlastic/empowerchain#1191


ba1ciu:
@PiranavanShanmugavadivelu I'm reverting this PR, because it messed up the images on featured and auction pages. Needs to be looked into again.

PR #1209, opened 2024-03-03
Rename everything auction to listing or project if its user facing

jschill:
Rename everything auction to listing internally and if its something a user will see we call it projects. This will also rename the urls and fix filenames of the pages.


frontend/marketplace/src/router/index.ts
```diff
@@ -52,14 +50,14 @@ const router = createRouter({
       component: FAQPage,
     },
     {
-      path: "/auction",
+      path: "/listings",
```
jschill, line 53: Maybe /projects is more appropiate here

frontend/marketplace/src/router/index.ts
```diff
@@ -14,14 +14,12 @@ const HomePage = () => import("@/pages/HomePage.vue");
...
   import("@/pages/CertificatesAndCreditsPage.vue");
-const AuctionPage = () => import("@/pages/AuctionPage.vue");
-const AuctionDetails = () => import("@/pages/AuctionDetails.vue");
-const CallbackView = () => import("@/pages/CallbackView.vue");
-const UserProfile = () => import("@/pages/UserProfile.vue");
-const AuctionPaymentSuccessful = () =>
-  import("@/pages/AuctionPaymentSuccessful.vue");
-const AuctionPaymentCancelled = () =>
-  import("@/pages/AuctionPaymentCancelled.vue");
+const Listings = () => import("@/pages/ListingsPage.vue");
```
PiranavanShanmugavadivelu, line 17: It would be better if we use the same naming convention here 
`const ListingsPage = () => import("@/pages/ListingsPage.vue");`

PiranavanShanmugavadivelu (changes requested):

frontend/marketplace/src/router/index.ts
jschill: Fixed

PiranavanShanmugavadivelu (approved):

PR #1232, opened 2024-04-05
button enable issue fix

PiranavanShanmugavadivelu:
Issue - https://github.com/EmpowerPlastic/empower-platform/issues/6633


frontend/marketplace/src/components/CertificateHolderModal.vue
```diff
@@ -33,6 +33,8 @@ const isButtonDisabled = () => {
       isCertificateHolderValid.value &&
       isCertificateHolderEmailValid.value) ||
     (props.modalType === CertificateHolderModalTypeEnums.AUTHORIZED_USER &&
+      isCertificateHolderValid.value) ||
```
portlandnorway, line 36: This is starting to become `if` hell now, with 7 unreadable lines if you lack context. I would either split out each one of these `||` blocks into computed properties, like `isValidEmailOnlyUser` || `isValidAuthorizedUser` || `isValidCryptoUser`

or

make it more readable by extracting the isCertificateHolderValid property as it's needed every time.
```.javascript
const isButtonDisabled = () => {
  // always required
  if (!isCertificateHolderValid.value) {
    return true
  }

  return !(
    props.modalType === CertificateHolderModalTypeEnums.AUTHORIZED_USER ||
    props.modalType === CertificateHolderModalTypeEnums.CRYPTO_WALLET_USER ||
    props.modalType === CertificateHolderModalTypeEnums.UNAUTHORIZED_USER && isCertificateHolderEmailValid.value
  );
};

```

Btw what's the diff betweeen an "authorized" user and a crypto wallet user? I assume they're both authorized so authorized is probably a bad name. I would also say we should not be naming these email only users "unauthorized" when they are authorized by email. But that's outside of this task. Just sayin.

frontend/marketplace/src/types/CertificateHolderModalTypeEnums.ts
```diff
@@ -1,4 +1,5 @@
 export enum CertificateHolderModalTypeEnums {
   AUTHORIZED_USER = "AUTHORIZED_USER",
   UNAUTHORIZED_USER = "UNAUTHORIZED_USER",
+  CRYPTO_WALLET_USER = "CRYPTO_WALLET_USER",
```
portlandnorway, line 4: I think this is the first good name we have here, the other two are: too general and just plain wrong. An unauthorized user would never be able to do anything, yet, they are. IMO that's authorized. 

Again - may be outside the scope of this task, but i'd consider re-naming these enums to be representative of the type of user they represent. Unauthorized = Email Only, Authorized = whatever kind of user that is.. Empower user?

portlandnorway (approved):
Lacking context but i think i see what you're going for. Few comments on long if statements and variable naming.

frontend/marketplace/src/types/CertificateHolderModalTypeEnums.ts
PiranavanShanmugavadivelu: We allow users to purchase credits without logging in. In this case, I refer to them as `UNAUTHORIZED_USER` because they do not have a login but can still purchase credits.

portlandnorway: I would expect an unauthorized user to be locked out, thrown errors at, etc. Not be able to purchase credits.

PiranavanShanmugavadivelu: > I would expect an unauthorized user to be locked out, thrown errors at, etc. Not be able to purchase credits. What about 'GUEST' or 'EMAIL_ONLY'?

I agree with `GUEST` 😄

frontend/marketplace/src/components/CertificateHolderModal.vue
portlandnorway: Yes that's nice 👍

frontend/marketplace/src/types/CertificateHolderModalTypeEnums.ts
PiranavanShanmugavadivelu: @portlandnorway , Can you have a look and resolve the conversations?

portlandnorway: @PiranavanShanmugavadivelu done

ba1ciu (approved):

PR #1239, opened 2024-04-17
Hide see certs btn if guest

portlandnorway:
Confirmed working for guests.

Could not confirm in local for email login, stuck in broken logto callback

[Link](https://www.awesomescreenshot.com/video/26803688?key=7c3639dca463934a443e473993826491)

Could not confirm in local for wallet login, no test crpyto to purchase with. 🤷 

Operating a little blind here, if someone has a setup to test wallet/email log-in local that would be 👌


ba1ciu (approved):
Tested all ✅

